PR #5, opened 2017-07-22
Dev

15chrjef:
- add animations
- add invest page
- add react pie tracker package
- implement  pie tracker


igorbarinov:
Hi @15chrjef . Thank you for the PR!

When I press " New crowdsale" on the home page I'm routed to 3rd page.
When I press "Choose contract" I'm forwarded to 1st page.

"New crowdsale" should open page 1
"Choose contract" will be hidden when we don't have any contracts signed with currently open ethereum account in Metamask. If there is a crowdsale contract, created by ICO wizard, the page will show a pop-up (not designed yet)

Please let me know what do you think.

15chrjef:
Fixed the routing error. Button has been commented out until we have a use case.

PR #97, opened 2017-09-02
(feature) implement pdf with data from step 4

15chrjef:
-implement pdf functionality download from step4

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,6 +21,7 @@
     "ethereumjs-abi": "^0.6.4",
     "install": "^0.10.1",
     "jquery": "^3.2.1",
+    "jspdf": "^1.3.3",
     "npm": "^5.3.0",
     "oracles-combine-solidity": "file:src/components/oracles-combine-solidity",
     "path": "^0.12.7",
--- a/src/components/stepFour.js
+++ b/src/components/stepFour.js
@@ -3,7 +3,7 @@ import '../assets/stylesheets/application.css';
 import { deployContract, getWeb3, getNetworkVersion, addWhiteList, setFinalizeAgent, approve, setTransferAgent, setMintAgent, setReleaseAgent, updateJoinedCrowdsales, transferOwnership, setReservedTokensListMultiple } from '../utils/web3'
 import { noMetaMaskAlert } from '../utils/alerts'
 import { defaultState, PDF_CONTENTS } from '../utils/constants'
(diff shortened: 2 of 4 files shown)


vbaranov:
@15chrjef there are some issues:

1. There is no compiler version in the pdf file
2. The source code is not fully displayed in the file. Only that part, that is visible from the page
3. There is no ABI-encoded parameters in the file.
3. ABI of the contract looks like:
screen shot 2017-09-03 at 12 27 37

15chrjef:
@vbaranov 
1. The compiler version is in the smart contract ' pragma solidity ^0.4.11;'  Would you prefer that the hard coded string of the compiler version is added as a separate field to the pdf?
2. Will implement
3. Where are the parameters and what is the path to access the parameters via this.state?
4. Will implement

vbaranov:
@15chrjef 
1. Compiler version could be get from here. 
screen shot 2017-09-03 at 12 46 25
I think, we could make a constant, that will displayed in the page and in the pdf file. 
In the future, we'll automatically set actual compiler version to this constant at deploy stage.

3. this.state.contracts[i].abiConstructor
screen shot 2017-09-03 at 12 47 07

15chrjef:
Great. I'll get to both those items today.

vbaranov:
@15chrjef 
1) it works
2) A part of the source code is hidden at the bottom in every page in pdf file:
screen shot 2017-09-04 at 01 14 57
 
3) Please, remove brackets from value of ABI-encoded constructor arguments:
screen shot 2017-09-04 at 01 15 11


4) The same as for the source code in 2) : ABI isn't fully displayed
screen shot 2017-09-04 at 01 15 06

15chrjef:
@vbaranov 

2. Every part of the source contact is visible on my end. What steps did you take for contract creation?
3. done
4. same as 2.

*I added more space for 2. and 4. but the result is the same, I saw no cut off code before or after the extra space I added.*

vbaranov:
@15chrjef 
The same issue for 2) and 4). It is not depend on parameters of previous steps.
Please, see attached pdf.
[crowdsale (2).pdf](https://github.com/oraclesorg/ico-wizard/files/1275280/crowdsale.2.pdf)

15chrjef:
@vbaranov I don't have this issue. What type of contract are you using? What fields are you submitting?
[crowdsale (1).pdf](https://github.com/oraclesorg/ico-wizard/files/1275379/crowdsale.1.pdf)

PR #199, opened 2017-09-27
(Feature) Payment proccess views

Dobriy-i:
localhost-3000-invest


igorbarinov:
@Dobriy-i Igor, we have english speaking contributors, please use English 🙂

Dobriy-i:
Oh sorry :) i have made payment's static states, can anyone develop it next?

igorbarinov:
Sure, @vbaranov will take care of it. Thank you!

15chrjef:
@Dobriy-i Please pull the latest changes from the master branch

Dobriy-i:
@15chrjef done

15chrjef:
Thank @Dobriy-i can you also detail the purpose of this PR via the format here https://github.com/oraclesorg/ico-wizard/wiki/Pull-Request-Policy ?

fvictorio:
What should happen when you select Kovan instead of MetaMask? The only thing I can think of is ignoring the injected `web3` and connecting directly with Kovan (although I'm not sure how to do that).

igorbarinov:
@fvictorio I've added description of this PR here https://github.com/oraclesorg/ico-wizard/issues/319

fvictorio:
I comitted a change that should move things towards #319. Right now, it allows selecting the type of investement (MetaMask or QR). If QR is selected, the address is shown and you can copy it by clicking it.

There are several pending things, but I think this can be merged and the rest of the steps can be done in individual steps, unless you prefer a different workflow.

Pending things:
[ ] Show QR with the crowdsale address
[ ] Make sure the MethodID is the correct one (or even better: get it dynamically)
[ ] Disable the MetaMask option if web3 is not injected
[ ] Give some kind of feedback when the address is copied
[ ] Detect that the payment was done (although I don't know if this is possible if web3 is injected, unless you use some HTTP server that does this)

src/components/invest/index.js
```diff
@@ -273,6 +276,42 @@ export class Invest extends React.Component {
     return { days, hours, minutes}
   }
 
+  renderInvestThroughQR() {
+    return (
```
15chrjef: I would extract all of this to a separate React component.

src/components/invest/index.js
```diff
@@ -370,11 +410,20 @@ export class Invest extends React.Component {
...
+                <option value="qr">QR</option>
+              </select>
+              <a className="button button_fill" onClick={this.investToTokens}>Invest</a>
+            </div>
             <p className="description">
             Think twice before investment in ICOs. Tokens will be deposited on a wallet you used to buy tokens.
             </p>
           </form>
+          {
+            this.state.investThrough === 'qr' ? this.renderInvestThroughQR() : null
```
15chrjef: extract to variable

fvictorio:
@15chrjef Changes applied.

src/components/invest/InvestThroughQR.js
```diff
@@ -0,0 +1,39 @@
+import React from 'react'
+import { CopyToClipboard } from 'react-copy-to-clipboard';
+
+export default ({ crowdsaleAddress }) => {
```
15chrjef: the component being exported by default should have the same naming as the file that contains it. In this case 'InvestThroughQR', but it may be more appropriately named QRPaymentProcess

src/components/invest/index.js
```diff
@@ -386,17 +390,27 @@ export class Invest extends React.Component {
...
+                <option value={INVESTMENT_OPTIONS.qr}>QR</option>
+              </select>
+              <a className="button button_fill" onClick={this.investToTokens}>Invest</a>
+            </div>
             <p className="description">
             Think twice before investment in ICOs. Tokens will be deposited on a wallet you used to buy tokens.
             </p>
           </form>
+          {
+            this.state.investThrough === INVESTMENT_OPTIONS.qr ?
```
15chrjef: please extract 
```
this.state.investThrough === INVESTMENT_OPTIONS.qr ? 
<InvestThroughQR crowdsaleAddress={this.state.crowdsaleAddress} /> :  null
```
to a variable and render that variable

src/utils/constants.js
```diff
@@ -177,3 +177,8 @@ export const FILE_CONTENTS = [
 
 export const DOWNLOAD_NAME = 'contractInfo'
 export const DOWNLOAD_TYPE = 'text/plain'
+
+export const INVESTMENT_OPTIONS = {
+  metamask: 'metamask',
```
15chrjef: please make any property of a constant capitalized. For example
```
{
   METAMASK: 'metamask',
   QR:'qr'
}
```

src/components/invest/index.js
fvictorio: As far as I know, you can't do that in a template, can you? I mean, the code inside the template must be an expression, so you can use assignments.

The alternative is to move the expression to a function, and do:

```
renderQRPaymentProcess() {

{ this.renderQRPaymentProcess }
```

fvictorio:
Thanks for the feedback, @15chrjef. I committed two of them. I have some reserves about the third one; I explained what I think below that comment.

src/components/invest/index.js
15chrjef: One can use either return the template in a function or assign it to a variable. Either way will work.

15chrjef:
@fvictorio Thank you for addressing those. Let me know if we need consensus on any additional items.

fvictorio:
@15chrjef Changes applied. There is an issue with the "Contribute now" button: since its text is longer than before, it shows under the select (at least in my screen). But I think that fix shouldn't be part of this PR.

15chrjef:
Thanks. LGTM!

fvictorio:
@vbaranov Personally, I prefer an approach of smaller and more short-lived PRs. I think those three things can be done in separate PRs. Of course, I'm OK with doing them here, but I think is more productive to do them separately. Smaller PRs are easier to review and reduce the possibility of conflicts with other branches. What do you think?

igorbarinov:
I do agree with @fvictorio on short PRs. This one PR was introduced before our PR policy and in future let's try to stick with smaller PRs.

igorbarinov:
Let's remove "Send up to N ETH to this address" (now it's 15) when there is no max cap?

fvictorio:
I don't see an option to set a max cap, only min cap, so I removed the message entirely.

vbaranov:
@fvictorio maxCap is from white list

fvictorio:
I'll check that, thanks.

fvictorio:
I looked into this and, while there is a `whitelist` array with information of the min/max caps for each whitelisted addresses, that information is only available if you just created the crowdsale (is preserved from step 3). But if you refresh, or if you bookmarked the investment page on opened it in a new tab, the whitelist information is not available.

I guess this information is retrievable, but it seems like a bigger change. I suggest opening an issue for this (something like: show min/max cap if the user is connected with one of the whitelisted addresses) and working on that in a separate PR. What do you think?

vbaranov:
>I guess this information is retrievable, but it seems like a bigger change. I suggest opening an issue for this (something like: show min/max cap if the user is connected with one of the whitelisted addresses) and working on that in a separate PR. What do you think?

I agree to do it separately. What stops me to approve this PR is:

image 2017-11-14 19-00-04

Maybe, let's do like this? (I've added these changes to PR) :
screen shot 2017-11-14 at 19 04 47
@igorbarinov what do you think?

PR #226, opened 2017-10-05
Fix doubled callback

chebykin:
Found that an each contract creation callback is invoked 2 times due the logic of `web3.js` library:
- first time without `contract.address` assigned, right after transaction was sent
- second time with `contract.address` assigned, after a contract was created

Suggest to invoke `deployed` callback after contract creation only, so we'll have an access to it's address.

### DIFF
--- a/src/utils/blockchainHelpers.js
+++ b/src/utils/blockchainHelpers.js
@@ -213,10 +213,7 @@ export function deployContract(i, web3, abi, bin, params, state, cb) {
           }
           else if (contract.transactionHash) {
             checkTxMined(web3, contract.transactionHash, function txMinedCallback(receipt) {
-              if (receipt) {
-                if (receipt.blockNumber)
-                  return cb(null, receipt.contractAddress);
-              } else {
+              if (!receipt) {
                 setTimeout(checkTxMined(web3, contract.transactionHash, txMinedCallback), 1000);
               }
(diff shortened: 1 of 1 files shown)


igorbarinov:
Thank you @chebykin for contribution.
@vbaranov will review it.

Btw, do you have any suggestions/requirements

chebykin:
Thank you for this great tool. Hope to see it will gain popularity soon

I've just checked out the crowdsale configuration steps only. What I have noticed, that it will be nice to see some progress visualization on the contracts deployment step. There are at least 5 contracts to deploy and the same screen all the time with no any update. Those ones who don't read documentation will thought that they are stuck with an infinete loop or something is broken.

Also, due the absence of the progress it was hard to notice that with the doubled callback invocation I had to deploy about 30 contracts.

I see a huge refactoring work being done in another PR and probably these changes are already there ;)

vbaranov:
This bug should be solved here: https://wizard-stage.oracles.org/.
I have merged changes to [stage](https://github.com/oraclesorg/ico-wizard/tree/stage) branch.
@chebykin please, don't stop. We wait for new contributions from you!

chebykin:
Oh, seems I haven't tested it well. Thank you for pointing this out. I'm closing this issue and will watch on the new one.

PR #276, opened 2017-10-17
Fix: price of 1 token in wei > 18 decimals

vbaranov:
Related to https://github.com/oraclesorg/ico-wizard/issues/274
This issue blocks crowdsale creation at step 4 with specific rates.
Solution:
*roundToDecimals* function is introduced in global utils and it is used to round price of 1 token in wei to 18 decimals.


src/utils/utils.js
```diff
@@ -281,7 +281,9 @@ export const getconstructorParams = (abiConstructor, state, vals, crowdsaleNum)
                 case "_oneTokenInWei": {
                   //params.vals.push(state.pricingStrategy[crowdsaleNum].rate);
                   //params.vals.push(state.web3.toWei(1/state.pricingStrategy[crowdsaleNum].rate/10**state.token.decimals, "ether"));
-                  params.vals.push(state.web3.utils.toWei(1/state.pricingStrategy[crowdsaleNum].rate, "ether"));
+                  
+                  let oneTokenInETH = roundToDecimals(1000000000000000000, 1/state.pricingStrategy[crowdsaleNum].rate)
```
15chrjef: Extract the 1/state.pricingStrategy[crowdsaleNum].rate to a variable or function call

src/utils/utils.js
```diff
@@ -281,7 +281,9 @@ export const getconstructorParams = (abiConstructor, state, vals, crowdsaleNum)
...
+                  params.vals.push(state.web3.utils.toWei(oneTokenInETH, "ether"));
```
15chrjef: extract 'state.web3.utils.toWei(oneTokenInETH, "ether")' to  a function call or variable

src/utils/utils.js
```diff
@@ -292,6 +294,12 @@ export const getconstructorParams = (abiConstructor, state, vals, crowdsaleNum)
     return params;
 }
 
+export function roundToDecimals(n, input) {
+  let output = input
```
15chrjef: instead return 1.0 / n * Math.ceil(n * output)

src/utils/utils.js
```diff
@@ -281,7 +281,10 @@ export const getconstructorParams = (abiConstructor, state, vals, crowdsaleNum)
                 case "_oneTokenInWei": {
                   //params.vals.push(state.pricingStrategy[crowdsaleNum].rate);
                   //params.vals.push(state.web3.toWei(1/state.pricingStrategy[crowdsaleNum].rate/10**state.token.decimals, "ether"));
-                  params.vals.push(state.web3.utils.toWei(1/state.pricingStrategy[crowdsaleNum].rate, "ether"));
+                  
+                  let oneTokenInETHRaw = 1/state.pricingStrategy[crowdsaleNum].rate
+                  let oneTokenInETH = roundToDecimals(1000000000000000000, oneTokenInETHRaw)
+                  params.vals.push(state.web3.utils.toWei(oneTokenInETH, "ether"));
```
15chrjef: I would extract 'state.web3.utils.toWei(oneTokenInETH, "ether")' to oneTokenInWei as well

src/utils/utils.js
```diff
@@ -292,6 +295,10 @@ export const getconstructorParams = (abiConstructor, state, vals, crowdsaleNum)
     return params;
 }
 
+export function roundToDecimals(n, input) {
```
15chrjef: 1) Use an es6 arrow function not a javascript function
2) Since the return value can be held in one line I would re-write it as 
```
export const roundToDecimals = (n, input) => 1.0 / n * Math.ceil(n * input)
```

PR #281, opened 2017-10-18
(Fix) floorToDecimals function updating to cover all use cases

vbaranov:
Related to https://github.com/oraclesorg/ico-wizard/issues/274
Function *roundToDecimals*, which truncates decimals to max 18,  is improved to cover all use cases.


igorbarinov:
@15chrjef please take a look at the PR

src/components/stepFour/index.js
```diff
@@ -267,7 +267,7 @@ export class stepFour extends stepTwo {
   //FlatPricing
   getPricingStrategyParams = (web3, pricingStrategy, i, token) => {
     console.log(pricingStrategy);
-    let oneTokenInETH = roundToDecimals(1000000000000000000, 1/pricingStrategy.rate)
+    let oneTokenInETH = roundToDecimals(-18, 1/pricingStrategy.rate)
```
15chrjef: what is -18? We should not have plain strings or numbers, these values should all be extracted to constants and imported. I would also make '1/pricingStrategy.rate' a variable.

src/utils/utils.js
```diff
@@ -283,7 +283,7 @@ export const getconstructorParams = (abiConstructor, state, vals, crowdsaleNum)
                   //params.vals.push(state.web3.toWei(1/state.pricingStrategy[crowdsaleNum].rate/10**state.token.decimals, "ether"));
                   
                   let oneTokenInETHRaw = 1/state.pricingStrategy[crowdsaleNum].rate
-                  let oneTokenInETH = roundToDecimals(1000000000000000000, oneTokenInETHRaw)
+                  let oneTokenInETH = roundToDecimals(-18, oneTokenInETHRaw)
```
15chrjef: same as above for -18

src/utils/utils.js
```diff
@@ -295,8 +295,25 @@ export const getconstructorParams = (abiConstructor, state, vals, crowdsaleNum)
     return params;
 }
 
-export function roundToDecimals(n, input) {
-  return 1.0 / n * Math.ceil(n * input)
+export const roundToDecimals = (n, input) => Math.floor10(input, n)
+
+const decimalAdjust = (type, value, exp) => {
+    if (typeof exp === 'undefined' || +exp === 0) {
```
15chrjef: 1) why would exp be equal to the string 'undefined'? This should most likely be the value undefined.
2) what are you trying to do with +exp? if its number coersion, why not make it a number when it is passed in initially?
3) what does having an undefined exp or an exp equal to 0 mean in words? The logic contained within the if conditional should be extracted to a variable as well.

src/utils/utils.js
```diff
@@ -295,8 +295,25 @@ export const getconstructorParams = (abiConstructor, state, vals, crowdsaleNum)
...
+      return Math[type](value);
```
15chrjef: what is this trying to do here?

src/utils/utils.js
```diff
@@ -295,8 +295,25 @@ export const getconstructorParams = (abiConstructor, state, vals, crowdsaleNum)
...
+    }
+    value = +value;
```
15chrjef: why all the '+value's? There should be a clearer way of accomplishing your intent here.

src/utils/utils.js
```diff
@@ -295,8 +295,25 @@ export const getconstructorParams = (abiConstructor, state, vals, crowdsaleNum)
...
+    exp = +exp;
+    if (isNaN(value) || !(typeof exp === 'number' && exp % 1 === 0)) {
```
15chrjef: I would extract the clause in the if conditional to a variable as well as !(typeof exp === 'number' && exp % 1 === 0)

src/utils/utils.js
```diff
@@ -295,8 +295,25 @@ export const getconstructorParams = (abiConstructor, state, vals, crowdsaleNum)
...
+      return NaN;
+    }
+    value = value.toString().split('e');
```
15chrjef: Instead of these three variable assignments that are self referencing, build a util function to extract this logic and return the final value for 'value'

src/utils/utils.js
vbaranov: Let's use  Math.floor10 and decimalAdjust. It is well-known solution: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/round

src/utils/utils.js
vbaranov: Let's use  Math.floor10 and decimalAdjust. It is well-known solution: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/round

src/utils/utils.js
vbaranov: Let's use  Math.floor10 and decimalAdjust. It is well-known solution: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/round

src/utils/utils.js
vbaranov: Let's use  Math.floor10 and decimalAdjust. It is well-known solution: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/round

src/utils/utils.js
vbaranov: Let's use Math.floor10 and decimalAdjust. It is well-known solution: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/round

src/components/stepFour/utils.js
```diff
@@ -220,7 +221,7 @@ export function transferOwnership(web3, abi, addr, finalizeAgentAddr, gasLimit,
     }
     if (!tokenContract) return noContractAlert();
 
-    sendTXToContract(web3, tokenContract.methods.transferOwnership(finalizeAgentAddr).send({gasLimit: gasLimit, gasPrice: 21000000000}), cb);
+    sendTXToContract(web3, tokenContract.methods.transferOwnership(finalizeAgentAddr).send({gasLimit: gasLimit, gasPrice: GAS_PRICE}), cb);
```
15chrjef: What is 
```
tokenContract.methods.transferOwnership(finalizeAgentAddr).send({gasLimit: gasLimit, gasPrice: GAS_PRICE})
```
I would extract this to a variable

src/utils/utils.js
15chrjef: I agree. All the '+' signs in the expressions will be replaced by Math.round? Or other methods on the Math object?

src/utils/utils.js
15chrjef: Sounds good. Please extract line 306 to a variable

src/utils/utils.js
15chrjef: sure. Can you rename 'value' to something more descriptive?

vbaranov: value -> inputNumber

src/utils/utils.js
vbaranov: Ok, done

src/utils/utils.js
vbaranov: it is the numeric representation: https://stackoverflow.com/questions/6682997/javascript-plus-symbol-before-variable

src/components/stepFour/utils.js
vbaranov: ok, done

vbaranov:
@15chrjef, is it time to approve?

src/utils/utils.js
```diff
@@ -295,8 +295,25 @@ export const getconstructorParams = (abiConstructor, state, vals, crowdsaleNum)
     return params;
 }
 
-export function roundToDecimals(n, input) {
-  return 1.0 / n * Math.ceil(n * input)
+export const roundToDecimals = (n, input) => Math.floor10(input, n)
+
+const decimalAdjust = (type, value, exp) => {
```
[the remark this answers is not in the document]
vbaranov: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/round

PR #306, opened 2017-10-25
(Fix) Use Oracles favicon

fvictorio:


fvictorio:
@vbaranov I updated this. Does it need some other change?

vbaranov:
No, thanks. Everything is fine now.

PR #325, opened 2017-11-01
(Fix) Use web3 HttpProvider in dev environment

fvictorio:
Depends on #324.

If web3 is not injected and you are in the `development` environment (using `npm start`), try to connect directly to testrpc (`localhost:8545`).

I think this can be very useful during development, since the transactions are submitted automatically.


src/utils/blockchainHelpers.js
```diff
@@ -38,6 +38,10 @@ export function getWeb3(cb) {
 	if (typeof web3 === 'undefined') {
     // no web3, use fallback
     console.error("Please use a web3 browser");
+    if (process.env.NODE_ENV === 'development') {
```
15chrjef: extract to variable

fvictorio: Do you mean the condition in the `if`? Something like:

```js
const devEnvironment = process.env.NODE_ENV === 'development'
if (devEnvironment) {
```

15chrjef: yes exactly 👍

fvictorio:
@15chrjef Change applied.

PR #334, opened 2017-11-02
(Fix) No error message on invest form submit

### LINKED ISSUE #273: Add validation on Invest button
Problem: It is possible to press invest with an empty token field and get success message

image

If invest amount is undefined  show a validation message under the invest page "Number of tokens to buy should be positive "

fernandomg:
Closes #273 (https://github.com/oraclesorg/ico-wizard/issues/273#issuecomment-341449487)

- Displays error message if the user tries to Invest without entering a valid token amount.
- Moved input's pristine flag, into the Component's state.

error_message-appears_on_pristine_submit

### DIFF
--- a/src/components/invest/index.js
+++ b/src/components/invest/index.js
@@ -11,13 +11,13 @@ import { defaultState, GAS_PRICE } from '../../utils/constants'
 export class Invest extends React.Component {
   constructor(props) {
       super(props);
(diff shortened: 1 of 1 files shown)


vbaranov:
@fernandomg is this completed PR? Could you, please, move to `awaiting for review` label, when PR is ready to review?

PR #354, opened 2017-11-08
(Refactor) Add basic test

fvictorio:
I tried to add a test in another branch I'm working on and I found out that tests that depend (directly or indirectly) on `web3` would fail. This happens because the version of `react-scripts` we are using has an issue with files with `.node` extension (`web3` depends on `scrypt`, that has a dependency of this type). See [the last bug fix here](https://github.com/facebookincubator/create-react-app/blob/master/CHANGELOG.md#bug-bug-fix-5) for more information.

I added some basic tests to demonstrate the problem. If you checkout this branch and run `npm run test:dapp`, you'll see the problem. Upgrading the dependencies (`npm install`) should fix that issue.

The last version of `react-scripts` is `1.0.17` and there seems to be no breaking changes, but I just upgraded to `1.0.11` to avoid unnecessary risks.


15chrjef:
@fvictorio can you please update the same PR when going through review rather than creating a new PR with the proposed changes?

PR #358, opened 2017-11-09
(Feature) Validate addresses used in reserved tokens

fvictorio:
This feature is simple, but I had to do some serious refactor around some components. I'll describe here what I had to do and some anti-patterns that I think we should avoid.

1. Avoid mutating React state. Things like:

    ```js
    const state = this.state
    state.some.deep.object.var = 'foo'
    this.setState(state)
    ```

    should be avoided. Sadly, there is no easy alternative that I'm aware of. The underlying problem is that React components should not have state with deep objects, but that's harder to fix. The best thing I found was [immutability-helper](https://github.com/kolodny/immutability-helper). Using that, the previous example becomes:

    ```js
    const newState = update(this.state, {
      some: {
        deep: {
          object: {
            var: { $set: 'foo' }
          }
        }
      }
    })

    ```

    It has some issues IMO, but it works and it's [recommended](https://reactjs.org/docs/update.html) in React's official docs.

2. Use [stateless functional components](https://hackernoon.com/react-stateless-functional-components-nine-wins-you-might-have-overlooked-997b0d933dbc) when possible. They are easier to understand, easier to test and have better performance. You can see a stateful component being written as stateless in the `InputField` and `ReservedTokensItem` diffs in this PR.

3. Avoid [uncontrolled components](https://reactjs.org/docs/uncontrolled-components.html). This was the biggest issue I found when trying to implement this simple validation. The state was hard to find and the logic was too complex. Now there is only one uncontrolled component, `AlertContainer`, but that one is meant to be used that way, so it's OK.

4. Don't commit code used for debugging. I removed several `console.log`s because I felt they weren't useful and, if you wanted to add some temporary log, it was hard to find it between the noise. If someone was using the logs, please let me know and I'll undo those deletions, but I think logs should be temporary.

Sorry for the big PR and long explanation! Let me know if you disagree with any of these points. Hopefully we can reach a more manageable amount of technical debt and increase our velocity.

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,6 +25,7 @@
   },
   "dependencies": {
     "clipboard": "^1.7.1",
+    "immutability-helper": "^2.5.0",
     "install": "^0.10.1",
     "jquery": "^3.2.1",
     "mobx": "^3.3.0",
--- a/src/components/Common/CrowdsaleBlock.js
+++ b/src/components/Common/CrowdsaleBlock.js
@@ -2,7 +2,7 @@ import React from 'react'
 import '../../assets/stylesheets/application.css';
 import { WhitelistInputBlock } from './WhitelistInputBlock'
 import { defaultCompanyEndDate } from '../../utils/utils'
-import { InputField } from './InputField'
+import InputField from './InputField'
 import { RadioInputField } from './RadioInputField'
 import { VALIDATION_MESSAGES, VALIDATION_TYPES, TEXT_FIELDS } from '../../utils/constants'
 const { EMPTY, VALID, INVALID } = VALIDATION_TYPES
--- a/src/components/Common/InputField.js
+++ b/src/components/Common/InputField.js
@@ -1,49 +1,39 @@
-import React from 'react'
+import React from 'react';
 import '../../assets/stylesheets/application.css';
-import { VALIDATION_TYPES } from '../../utils/constants'
-const { INVALID } = VALIDATION_TYPES
+import {VALIDATION_TYPES} from '../../utils/constants';
 
-export class InputField extends React.Component {
-	constructor(props) {
-		super(props);
-		this.state = {};
-	}
+const {INVALID} = VALIDATION_TYPES;
 
-	clearVal = () => {
-		let state = {...this.state}
-		state.val = ""
-		this.setState(state)
-	}
+const InputField = props => {
+  const errorStyle = {
+    color: 'red',
+    fontWeight: 'bold',
+    fontSize: '12px',
+    width: '100%',
+    height: '10px',
+  };
 
-	onChange = (e) => {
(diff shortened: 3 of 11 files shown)


fvictorio:
Oh, that's strange. I'll test it again, thanks.

fvictorio:
@vbaranov  I tested it with MetaMask and testrpc and it seems to work. What did you use?

vbaranov:
I've tested with testrpc. I added reserved tokens in perecents and in absolute value for the same address. 10 for both. Crowdsale and invest pages were broken: there is no any information except address of crowdsale. I tested twice.

fvictorio:
Mmh, I'm not being able to reproduce it. This is what I'm doing: https://youtu.be/HcW83qT7Qic

vbaranov:
@fvictorio it was the problem with gas in testrpc. Deployment works, but reserved tokens functionality is still broken: value in tokens is empty. See, please, example:

screen shot 2017-11-10 at 18 17 26
screen shot 2017-11-10 at 18 22 40

15chrjef:
@fvictorio yes our state management is not where it should be. I have refactored using mobx. You can find it here. 

https://github.com/15chrjef/ico-wizard/pull/4

fvictorio:
@vbaranov I think it's fixed now. But I spot a problem: the tokens you reserve seem to be multiplied by ten (if you reserve 2 tokens, 20 appear reserved). I reproduced the problem in `master`, so I don't think it's related with this PR. If you can confirm is an issue, I can work on it on a different PR.

vbaranov:
@fvictorio I haven't seen mentioned problem before. Could you give the link to the token contract, created from master, where the problem exists?

fvictorio:
Ok, I investigated this a little and I found what's happening. Not sure if it is a bug or an implementation detail.

If you reserve 5 tokens, and input 1 decimal of precision, then the reserved tokens list shows "50". If you input 2, then it shows "500". It's like it's showing the amount of the minimal unit of tokens that is buyable.

Is this something that should be fixed or just the way that decimals are implemented in the contracts?

vbaranov:
@fvictorio Are you talking about `inTokens` property of token contract? `inTokens` property should store this value: amount_of_resereved_tokens_in_absolute_value * token_decimals

fvictorio:
Oh, yes, sorry for not being clear about that. I was talking about the same property that in the screenshot of  the contract values. That is working fine, then.

vbaranov:
Problem: Input fields for 2nd and next tiers are disabled, date fields are empty.
I reverted this PR: https://github.com/oraclesorg/ico-wizard/pull/358
I restored `validate-addresses` branch
The original issue is here: https://github.com/oraclesorg/ico-wizard/issues/285
@fernandomg @fvictorio could you take a look?

Issue #292, opened 2017-10-22
Byzantium supports success/fail tx


rstormsf:
https://ethereum.stackexchange.com/questions/28077/how-do-i-detect-a-failed-transaction-after-the-byzantium-fork-as-the-revert-opco/28078?stw=2#28078

every tx receipt is now has status:
`status: "0x1",` - success
` status: "0x0",` - fail

if it was called by `revert`

fernandomg:
So far this is doable, but MetaMask is having issues sending the error to the Dapp (https://github.com/MetaMask/metamask-extension/issues/2060) so the Dapp never gets a response from the contract method call.

In the other hand, I've been trying with Mist but I'm experiencing this issue: https://github.com/ethereum/mist/issues/2131.

So I'd wait for those problems to be fixed in order to continue with this issue, otherwise is too hard to develop and test.

rstormsf:
what issues? why do u use testrpc? testrpc is not real node, so don't expect testrpc to support it

fvictorio:
testrpc does support Byzantium features: https://github.com/ethereumjs/testrpc/issues/402

What do you use for development if not testrpc?

PR #364, opened 2017-11-16
(Feature) Support for byzantium tx success/fail statuses

fernandomg:
This closes issue #292 

- For the invest page, toaster error message will always be the same: 'Transaction has failed, please retry' (if the user rejects a transaction, if a transaction ran out of gas or any other exception)
- It can be tested locally with testrpc ^6.0.0, it supports Byzantium features
- To test properly, block auto-mining should be disabled. That is: `testrpc --blocktime 1`, if not... no event will be triggered for on-chain errors.

### DIFF
--- a/src/utils/blockchainHelpers.js
+++ b/src/utils/blockchainHelpers.js
@@ -213,49 +213,17 @@ export function deployContract(i, web3, abi, bin, params, state, cb) {
   });
 }
 
-export function sendTXToContract(web3, method, cb) {
-  let isMined = false;
+export function sendTXToContract(method, cb) {
   method
-  .on('error', function(error) {
-    console.log(error);
-    return cb(error);
-  })
-  .on('transactionHash', function(transactionHash){
-    console.log("contract method transaction: " + transactionHash);
-
-    checkTxMined(web3, transactionHash, function txMinedCallback(receipt) {
-      if (isMined) return;
-
-      if (receipt) {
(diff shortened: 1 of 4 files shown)


src/utils/blockchainHelpers.js
```diff
@@ -213,49 +213,17 @@ export function deployContract(i, web3, abi, bin, params, state, cb) {
...
+export function sendTXToContract(method, cb) {
   method
-  .on('error', function(error) {
-    console.log(error);
-    return cb(error);
-  })
-  .on('transactionHash', function(transactionHash){
-    console.log("contract method transaction: " + transactionHash);
-
-    checkTxMined(web3, transactionHash, function txMinedCallback(receipt) {
```
vbaranov: This additional polling of tx receipt was made, because users had problems on mainnet: wizard hanged on random transaction, because there wasn't response from it, no receipt. Especially, if you switch between tabs when wizard works. Did byzantium hard fork fix it? I doubt, that we need to change this part without real need, even if hand tests will successfull on mainnet.

fernandomg: Ok, good to know. I'll rollback the changes, and will leave a comment with that explanation for future references. Thanks for the input!

fernandomg:
@vbaranov... created an issue to not loose track of that toast suggestion: https://github.com/oraclesorg/ico-wizard/issues/367

PR #378, opened 2017-11-29
(Feature) Validate address

fvictorio:
This re-adds the address validation. It also adds it in the whitelist component (the previous PR only validated the reserved tokens component).


vbaranov:
This PR also fixes the problem, when the white list form doesn't appear for the 2nd tier, if we enable it after we added 2nd tier form.

PR #381, opened 2017-12-01
(Chore) Add issue template

fvictorio:
There have been several reported issues that don't provide some basic information. This adds a template so that people opening bugs can know which is the minimal data that they should include with the report.


igorbarinov:
Great idea, Let's add a memo how to add log from google chrome dev console

fvictorio:
Updated

igorbarinov:
Thank you

PR #382, opened 2017-12-01
(Chore) Use cross-env

### LINKED ISSUE #317: (Suggestion) Use cross-env to avoid duplicating NPM scripts
Right now there are several duplicated NPM scripts with the suffix `Win` and using windows-style environment variables.

An alternative way to handle this is to add [cross-env](https://github.com/kentcdodds/cross-env) as a devDependency and replace

```
scriptName: node $someVar main.js,
scriptNameWin: node %someVar% main.js
```

with

```
scriptName: cross-env node $someVar main.js
```

fvictorio:
Closes #317.

I don't have Windows so I couldn't test it there, but it works on Linux.


fvictorio:
I tested this in windows and it works, but it has a problem generating the contracts in `public/contracts`. As far as I can tell, the same issue happens in master, but since I can't really test that everything is working, I'm going to close this PR and the issue.

If someone wants they can give it a try, but since none of us seems to use windows for developing, I think this is really low priority for now.

PR #391, opened 2017-12-06
(Feature) Add options to select gas price

### LINKED ISSUE #283: Gas price optimizations and UI improvements
Problem: After Byzantinium hard fork gas price reduced dramatically for average transactions
At the moment the median gas price is 15Gwei versus 21Gwei.

>3) median gas price has fallen below 20 gwei since I have been tracking. Now at 15 gwei. Over 15% > using 1 gwei or less
https://twitter.com/ETHGasStation/status/920774738418417664

> 2) wait times are falling with 1 gwei gas price getting confirmed in under 1 minute (except for very high gas transactions)
https://twitter.com/ETHGasStation/status/920774470402416640

> 1) I) Big shifts going on in the ETH gas market post Byzantium- 20% of recent blocks include transactions with 0.1 gwei gas price, 40% w 1 gwei
https://twitter.com/ETHGasStation/status/920774071117156352

More details:
http://ethgasstation.info/

Solution:

a. Add list box with "Gas price" to "Global limits"
Description of the field "Slow is cheap, Fast is expensive. Check out gas price at 
- Slow and Cheap (1Gwei)
- Normal (5Gwei)
- Fast and Expensive(15Gwei)

b. Rename "Global limits" to "Global settings"
c.  Use selected price model for signing txs in metamask AND in alter message before submitting txs

### LINKED ISSUE #390: (Fix) Change gas price to 10 GWei
https://ethgasstation.info/
screen shot 2017-12-05 at 11 45 51

fvictorio:
Closes #283 and #390.


vbaranov:
@fvictorio merge conflicts need to be fixed here

fvictorio:
Done.

fvictorio:
I agree, but I'd propose something a little different.

First, the global settings should be in a separate card. Right now they appear in the first card (the first tier) and don't show up in the following tiers. I would create a separate card with the global settings (wallet address, whitelisting, min cap and gas price). This card would appear at the top, and then there would be a new card for each tier added.

I think that would be cleaner: all the tiers would have the inputs in the same position. And having the global settings visually separated make their intention clearer.

Also, I wouldn't hide the Min cap when whitelist is enabled, I would disable it, but that's a personal preference.

What do you think?

igorbarinov:
> First, the global settings should be in a separate card.

Good idea

> Also, I wouldn't hide the Min cap when whitelist is enabled, I would disable it, but that's a personal preference.

Good idea

vbaranov:
@igorbarinov Agree with these points. @fvictorio Could we visually emphasize disabling? By changing background of the input, for example.

fvictorio:
@vbaranov Sure!

fvictorio:
This should be ready for review. I left the prices at 5/10/15 Gwei.

igorbarinov:
Franco, by the way, can we deploy it on stage? or I'm supposed to review it on a local machine?
what is the best way to do the review of UI behavior

fvictorio:
I always test things locally, but it would be good to have a better process for this. What do you think of something like this?

1. PRs are reviewed locally by devs.
2. If they are approved, they got merged to stage, since that branch is already integrated with the CI.
3. Then someone is in charge of testing things in stage, and if everything works there, stage is merged in master. That would be our deploy to production.

It would mean that things get to production more slowly, but it would be safer. If we do this, though, we should add a "stage" column in waffle, I think.

Thoughts?

youtweakit:
It works. Suggest adding Custom Gas Price ;-)

igorbarinov:
@youtweakit good idea.
@fvictorio I hope we can set up an oracle for gas price on mainnet (very volatile gas price at the moment) using this code https://github.com/oraclesorg/ico-wizard/issues/393#issuecomment-350508366

Although, let add users to specify custom gas price

igorbarinov:
About staging I agree that we need it only for external testing 
or if we can make CD for it, e.g. every PR will be deployed on a stage server. Let me see if we have this functionality in Netlify.

igorbarinov:
Nice, Netlify deploys all previews automatically

image

I added hooks to GitHub. Netlify should inform on build preview

igorbarinov:
image

fvictorio:
Nice, that will be very useful.

fvictorio:
This is ready for review. I tested it a little in Netlify's deploy preview, it's great to have that.

youtweakit:
https://clever-bartik-1e0c01.netlify.com/ - works fine! Great tweak ;-) will use this build for our pre sale project. Thanks!

vbaranov:
@fvictorio @igorbarinov changes are approved. Let me clarify: we won't change scheme of merging. Anyone can check deploy preview in Netlify before merging. I will merge it to master, right?

fvictorio:
@vbaranov Yes, I agree.

PR #404, opened 2017-12-15
(Feature) Register created contracts

fvictorio:
This PR adds the basic feature of registering the crowdsale address in a Registry contract. It depends on oraclesorg/ico#7.

There are several things needed after this, but I want to verify that this approach works.

I added a `.env.example` file. For this to work, you need to copy that to a `.env` file. This (not committing the actual env file) is standard practice, although arguably not necessary for an app that has only front. Still, I preferred to err on the side of security.

This adds an environment file that maps network ids to the address of the deployed Registry contract. This means that, before starting the app, this contract needs to be deployed. To make that setup easier during development, the address in the example is the address of the first deployed contract when you use `ganache-cli` in deterministic mode. I added a NPM script that starts ganache in this mode and also uses a predefined network id. I also added another NPM script that deploys the Registry contract and starts the app.

tl;dr, do this:

1. `cp .env.example .env`
2. In one terminal, execute `npm run ganache`
3. In other terminal, execute `npm run dev`

Let me know if you think this approach works.


fvictorio:
Oh, and @igorbarinov, if we go this route, we'll need to deploy the Registry to the testnets and the mainnet, and add the proper environment variable to Netlify. Since the Registry contract is not final, I would not deploy it to the mainnet yet, but we can begin to test it in the testnets. You can give me access to the Netlify dashboard if you want, otherwise I will paste here the environment variable that needs to be added (I'll deploy the Registry to Kovan and Rinkeby and paste the addresses here).

fvictorio:
I deployed the Registry contract in Rinkeby and Kovan and I'm going to test the Netlify preview deploy, so I modifies `.gitmodules` to temporarily use my fork of `ico`. I'll change it back after performing the tests and when that PR is merged.

fvictorio:
Tested the Netlify deploy preview. Worked fine in both Rinkeby and Kovan.

igorbarinov:
Please deploy also on Core (https://core.poa.network) and Sokol (https://sokol.poa.network)
For Sokol, you can get free coins here https://faucet-sokol.herokuapp.com/
send me your wallet and I'll send you some test coins

fvictorio:
I deployed the registry to Sokol and updated the `.env.example` and Netlify configuration. I also tested the deploy preview with Sokol and it worked.

I created an account in Core, the address is `0x8220D966abA3C328eBf32858CB3A5fe868e02167`, if you send me some coins I'll deploy the registry there and test it.

igorbarinov:
@fvictorio thanks, I sent. Please check balance

fvictorio:
Deployed in Core, worked fine, although Netlify's cache gave me some headaches.

Next step is to merge https://github.com/oraclesorg/ico/pull/7 and then update the `.gitmodules` in this PR.

fvictorio:
This had several changes since it was reviewed, so please test it again and take a look at the last commits.

This should be functional now. The main thing pending is filling the columns in the list of crowdsales (right now it only shows the addresses, it should show the token name, ticker, and creation date). But I think we can show the button in production anyway, what do you think?

vbaranov:
@fvictorio 
I'm facing with this issue on the main page (Ropsten network):

screen shot 2017-12-29 at 18 47 31
It seems Metamask isn't injected into the page properly. The error happened when I tried to refresh the page. I tried in Chrome incognito mode too.

>But I think we can show the button in production anyway, what do you think?

I agree with button and only addresses in production as a first step.

fvictorio:
@vbaranov  That doesn't happen to me when I connect to Ropsten. The Registry wasn't added in that chain (I just added it), but that wouldn't cause that error. Any hints on how to reproduce it?

vbaranov:
In my tests this happened not systematically: not every refresh caused this error.

vbaranov:
I’ll let you know, if I’ll find the conditions of appearance of this error

fernandomg:
> I've tried to change rate of created crowdsale with one tier. And I signed 3 transactions to the same contract, the same method
The only way I found to reproduce it (in local or with kovan/rinkeby) was double-clicking `Save` button. I just set the _loading_ overlay immediately when the button is clicked and wasn't able to re-reproduce.

> I changed the start time and successfully signed the transaction. And then I changed end time: there were 2 transactions: for changing of the end time and again for changing of start time. I guess we should force refresh /manage page after successful update.
I agree with the refresh. Done.

> What should be in «Crowdsale setup name»
It should display Tier's name. In master the code is already merged. I just merged master and it worked fine.

> Style suggestions
Done.

fvictorio:
@vbaranov The env variables are already configured in Netlify. You can check it in Settings -> Build & Deploy -> Build environment variables.

vbaranov:
I see, thanks

vbaranov:
@fvictorio but, I don't see the registry address for mainnet in env variable in Netlify. Is it already created?

vbaranov:
@fvictorio I've created and added to Netlify env variable registry for mainnet

Issue #323, opened 2017-11-01
 (Feature) Manage created crowdsale contracts


igorbarinov:
**Rationale**
Users create multiple contracts using the wizard, usually for testing purposes. They need:
- to return back to their contracts without bookmarking them
- have the ability to modify updatable tiers
- have the ability to finalized contracts

**Implementation** 
On creating the crowdsale page there is a new button "Choose contract". It appears only if there is a crowdsale contract created by the active and unlocked wallet. The page will poll user's wallet for previously created contracts by the wizard.
 
![image](https://user-images.githubusercontent.com/424628/32289722-c0ea47b6-bef4-11e7-837c-bde333327e59.png)

If the button is active a list of crowdsales appears. Crowdsale name should be replaced to contract address, the name of a token, ticker, and date of creation. 
![image](https://user-images.githubusercontent.com/424628/32299373-a6cd814a-bf12-11e7-9d05-97320af883e3.png)

Updates of tier:

All tiers of the crowdsale are displayed. If tier itself or fields of a tier are not updatable then values are read-only.

Finalize button:

There is a finalize button to finalize the crowdsale. Should show a popup with yes/no and warning after pressing it. 
Text:
"Are you sure to finalize the crowdsale? After finalization, it's not possible to buy tokens, all tokens will be movable, reserved tokens will be issued. Press Yes to finalize, and NO to close this dialog."

Whitelists:

List of current whitelisted addresses is shown. It's possible to add an address to whitelist. 
If whitelisted address added in the state  

![image](https://user-images.githubusercontent.com/424628/32299395-b206c0da-bf12-11e7-9c06-12682c78ebab.png)

igorbarinov:
open questions
- [ ] how to get ABI for the contract. should we ask for it before opening configuration page?

vbaranov:
ABI is static. But if we assume, that the interface of the crowdsale contract can be changed in the future releases, yes, we need to ask ABI before opening configuration page.

xirkle:
Do I need to create multiple smart contracts in order to see the "choose contract" button? How do I modify one existing smart contract before or after the start date? Thanks!

igorbarinov:
@xirkle this PR is only for future. At the moment, you have to call function directly
e.g. to change max cap please refer to this wiki page https://github.com/oraclesorg/ico-wizard/wiki/Change-max-cap-for-tier

fvictorio:
Some thoughts on this:

[...]

PR #417, opened 2017-12-26
(Feature) Manage created crowdsale contracts edit data

fernandomg:
This PR is the fourth step towards the completion of #323

- Ability to modify StartTime, EndTime, Rate and Supply for each updatable Tier.
- What you think about adding a Popup with information about the fields/Tier to be edited before continue with the transactions submissions? As each field edited will represent a different Tx, it would be good to give some sort of feedback to the user about what he/she will be accepting/paying for.
- Whitelist is not enabled fully functional, as I'm waiting for the https://github.com/poanetwork/ico/pull/11
- Tier name will be available once https://github.com/poanetwork/ico/pull/10 is available


src/components/Common/CrowdsalesList.js
```diff
@@ -29,20 +29,20 @@ export default class CrowdsalesList extends Component {
         <div className="container-fluid">
           <div className="table-row flex-table-header">
             <div className="text">Address</div>
-            <div className="sm-text">Token Name</div>
-            <div className="sm-text">Ticker</div>
-            <div className="sm-text">Date</div>
+            {/*<div className="sm-text">Token Name</div>*/}
```
vbaranov: @fernandomg is it temporarily commenting or should we delete these commented lines before merging to master?

fernandomg: it's temporary, until we can retrieve the required data for the "Choose contract" popup.

But for the sake of clarity, I can delete it and add it back when required.

vbaranov:
>What you think about adding a Popup with information about the fields/Tier to be edited before continue with the transactions submissions? As each field edited will represent a different Tx, it would be good to give some sort of feedback to the user about what he/she will be accepting/paying for.

Nice idea. It would be extremely user-friendly.

src/components/Common/CrowdsalesList.js
vbaranov: Oh, I see. It would be double work, if we keep in mind, that we should return it later. Ok, let's leave commented lines here.

vbaranov:
@fernandomg some unmerged conflicts. @fvictorio could you review after resolving?

src/utils/constants.js
```diff
@@ -111,6 +111,8 @@ export const VALIDATION_MESSAGES = {
   START_TIME: 'Please enter a valid date later than now',
   END_TIME: 'Please enter a valid date later than start time',
   MULTIPLE_TIERS_START_TIME: 'Please enter a valid date not less than the end time of the previous tier',
+  EDITED_END_TIME: 'Please enter a valid date later than start time and previous than start time of next tier',
+  EDITED_START_TIME: 'Please enter a valid date later than now, less than end time and less than the end time of the previous tier',
```
fvictorio: Should be `and later than the end time of the previous tier`.

src/components/manage/index.js
```diff
@@ -119,11 +119,45 @@ export class Manage extends Component {
     e.stopPropagation()
 
     if (!this.state.formPristine) {
-      console.log('continue with saving...')
-      return
-    }
+      const { crowdsaleStore, tierStore } = this.props
+      const updatableTiers = crowdsaleStore.selected.initialTiersValues.slice().filter(tier => tier.updatable)
```
fvictorio: `.slice()` is not necessary here, `.filter` does not mutate the underlying array.

src/stores/TierStore.js
```diff
@@ -86,6 +86,49 @@ class TierStore {
...
+        if (index < this.tiers.length - 1) {
+          lessThanNextStart = validateLaterOrEqualTime(this.tiers[index + 1].startTime, this.tiers[index][property])
+        }
+
+        this.validTiers[index][property] = lessThanNextStart && laterTime ? VALID : INVALID
+        return
+      case 'startTime':
+        let notLaterTime = true
+        const previousToEndTime = validateLaterTime(this.tiers[index].endTime, this.tiers[index][property])
+        const validTime = validateTime(this.tiers[index][property])
```
fvictorio: Why does this check that the time is valid but the previous block doesn't?

fernandomg: It could be checked, but is redundant as `endTime` is compared against `startTime`, so it's forced to be a _validTime_.

Anyway, if the idea is to check it for `endTime` too, then `validateTiers` action should be changed as well: https://github.com/poanetwork/ico-wizard/blob/72ffdc488dca905ea59f66ec76ae1625788ccac0/src/stores/TierStore.js#L82

src/components/manage/index.js
fernandomg: fixed

src/utils/constants.js
fernandomg: done

src/stores/TierStore.js
fvictorio: Nvm, if it was intended it's OK for me.

PR #419, opened 2017-12-27
(Feature) Load crowdsale addresses from Registry

fvictorio:
Depends on #404.

Use the Registry to obtain the addresses of the crowdsales created by the current user.


fvictorio:
I pushed these commits in #404, so I'm going to close this and we can merge only that one.

PR #430, opened 2018-01-02
(Feature) Use Oracle's API for Gas Price

### LINKED ISSUE #393: (Fix) Transaction underpriced

Mainnet

There is no crowdsale. Issue happens on the first transaction.

image

No

---

fernandomg:
Closes #393 

- Requests gas price on stepThree 
- Requests gas price for invest page
- If it's not able to reach the API's endpoint, it will display an alert with the message: _ICO Wizard wasn't able to request current Gas Prices from the blockchain, custom values will be used_


igorbarinov:
Let's rename `Slow` to `Safe` in UI and make it default. 
People usually look for `safe` option on  in https://ethgasstation.info/ 
With the current price of Ethereum, it makes difference to use the cheapest version

@vbaranov what do you think?

fernandomg:
@igorbarinov FYI, `invest` Tx uses the `Fast` price, shall I change it to `Safe` too?

vbaranov:
Yes, 'Safe' sounds better than 'Slow'. And I agree with setting to the cheapest transactions by default everywhere in Wizard.

fernandomg:
@igorbarinov @vbaranov changes done

fernandomg:
@vbaranov You're right. It's unnecessary to consume the API for the testsnets.
But, in the other hand, I don't see the drawback of use it for the different testnets other than the amount of request done to the https://gasprice.poa.network endpoint

rstormsf:
@fvictorio Looks great, however no tests :-( 
at least 1 test that goes thru main flow would really help

vbaranov:
@rstormsf we could create tests in the background for this feature in a separate PR.
@fernandomg could you create a separate issue for oracle tests?
Relates to #461

PR #434, opened 2018-01-03
(Fix) Invest gas price zero

### LINKED ISSUE #433: (Bug) Sometimes gas price is set to 0 when trying to contribute in the invest page

All

Not related

No

No

---

fernandomg:
Closes #433


src/components/invest/index.js
```diff
@@ -210,7 +210,13 @@ export class Invest extends React.Component {
     }
     console.log(opts)
 
-    sendTXToContract(web3, crowdsaleContract.methods.buy().send(opts))
+    crowdsaleContract.methods.buy().estimateGas(opts)
+      .then(estimatedGas => {
+        const estimatedGasMax = 362897
```
vbaranov: @fernandomg is `estimatedGasMax` for 1-tier crowdsale without whitelist? In this case, `estimatedGasMax` will be greater for multiple tiers contract with whitelist. Because investing is more expensive, if whitelist is enabled due to https://github.com/poanetwork/ico/blob/master/contracts/CrowdsaleExt.sol#L239 and even more expensive, if whitelist + multiple tiers are switched on due to https://github.com/poanetwork/ico/blob/master/contracts/CrowdsaleExt.sol#L286

fernandomg: oh, you're completely right! I'll review that. Thanks!

fernandomg: @vbaranov 
I've been testing this in kovan.

Created a 3-tier crowdsale (https://wizard.poa.network/invest?addr=0xcCf27643fa6C4FC844a8945b7c2F8bd562153649&networkID=42) (**BTW: there's a bug with the Balance, that I'm about to fix**)

And bought tokens with `0x90f8bf6a479f320ead074411a4b0e7944ea8c9c1` (first ganache account on deterministic mode) on every tier.

This are the results (max used gas was for 1.i. with `240139`):
1. First Tier: 
    1. https://kovan.etherscan.io/tx/0xee8e111470a5093936fe640f909cdc59e5d8fe91c1007049cafc5ccf0f42ae7b
2.  Second Tier:
    1. https://kovan.etherscan.io/tx/0xc8702bd87a1235abb2fd73f257513d33de3fe33f54ccfd880570487a016e93b1
    2. https://kovan.etherscan.io/tx/0x6fd572f2ee0de8a67e463998cf645e811c10f0a1377268368d7faa8e97bd9283
    3.  https://kovan.etherscan.io/tx/0x7b43354455701e4590561b0c7bb8c2ef6874468b2d5ac5caea1317bda2f8fc65
3. Third Tier: 
    1. https://kovan.etherscan.io/tx/0x6b42cfc8104f82d4c4a2b0b92c3dab1359290e0912c64e7f3f5f9c4cc65c9b69

Please, give me a hint if there's something I'm missing out.

vbaranov: Yes, it seems that max gas indeed should be for investing at the 1st tier due to https://github.com/poanetwork/ico/blob/master/contracts/CrowdsaleExt.sol#L294.
But the gas used for tx should depend on the number of tiers. I suggest taking gas estimation for `estimatedGasMax` as investing in the 1st tier of 10-tiers crowdsale with whitelist. 10 - is the max amount of tiers I have ever seen in the real use cases.

fernandomg: @vbaranov I tested locally with a 12-tiers crowdsale and the max gas used was `376086` for the first tier. If it's Ok, what I'll do is to set `estimatedGasMax` to the same value than https://github.com/poanetwork/ico-wizard/blob/40978914427610c53f3393abc78979ad243a01ee/src/utils/blockchainHelpers.js#L116

BTW: While I was creating the crowdsale, found 2 tx that went out of gas:
https://github.com/poanetwork/ico-wizard/blob/40978914427610c53f3393abc78979ad243a01ee/src/components/stepFour/utils.js#L132

and
https://github.com/poanetwork/ico-wizard/blob/40978914427610c53f3393abc78979ad243a01ee/src/components/stepFour/utils.js#L154

Let me know if I fix it in a different PR, or should I consider it as part of this research/fix.

vbaranov: >If it's Ok, what I'll do is to set estimatedGasMax to the same value

Sounds good

>Let me know if I fix it in a different PR, or should I consider it as part of this research/fix.

Let's separate to different PRs

fernandomg: @vbaranov ready

PR #487, opened 2018-01-12
(Fix) Warning alert mainnet

### LINKED ISSUE #398: (Fix) There is no warning alert before step 4 in Mainnet

Mainnet

No URL

Alert is described here: https://github.com/oraclesorg/ico-wizard/blob/master/src/utils/alerts.js#L96.
It notifies users, how many transactions they should sign and estimated cost. It should appear only for Mainnet. I don't see it.

No

---

fernandomg:
Closes #398 

- Informs the amount of Tx that is about to sign
- Informs the aprox ETH based on price selected


fernandomg:
@vbaranov it's a Tx for a list of addresses (could be a list of one or more) for each tier. The same happens for the reserved ones.

vbaranov:
Ah, right.

PR #498, opened 2018-01-16
(Fix) Use eth.getAccounts() instead of eth.accounts

fvictorio:
The invest page doesn't seem to be working because of this.

`web3.eth.accounts` has a different behavior in web3 1.0. Not sure why this was working before.


vbaranov:
It seems invest works with `from: undefined` too, because metamask window appears on investing. But strictly saying we should change `eth.accounts` -> `eth.getAccounts()`.

PR #502, opened 2018-01-18
(Feature) five token ticker support

### LINKED ISSUE #501: (Fix) Increase token symbol length to 5 in validation
Problem:
Validation limits to 3 letters in a symbol. There is no requirements for that in the standard. Users are asking for more

Solution:
increase symbol length up to 5

vbaranov:
Closes #501 

**Problem**: Current max length of the token ticker is three.
**Solution**: Increase it to five.


vbaranov:
@fernandomg thanks for the comment. I've updated combinations number.

PR #503, opened 2018-01-18
(Feature) Deployment progress feedback to the user

### LINKED ISSUE #464: (Feature) Give user information about crowdsale deployment status and provide the ability to resume
Due to the recent issues we were facing with the crowdsale deployment, I think that will be a good idea to add an informative list with each step and its status in `stepFour`

i.e.:
- [x] safeMathLibrary
- [x] token
- [x] pricingStrategy
- [ ] crowdsale (... processing)
    - [x] tier 1
    - [ ] tier 2 (... processing)
    - [ ] tier 3 (pending)
- [ ] crowdsaleAddress (pending)
- [ ] finalizeAgentABIArguments (pending)
- [ ] finalizeAgent (pending)

Also, add the ability to resume from where the process was halted.

Thus, if the user don't leave the screen or refresh it; he/she will be able to continue from where the deployment failed if for instance the Tx goes OOG, Metamask fails for any reason, and that kind of issues.

@igorbarinov @vbaranov l look forward to your comments.

fernandomg:
This is the first step towards the completion of #464 

Presents a list of the required Txs and its statuses during crowdsale deployment.
Please **check the steps' descriptions**.

image


vbaranov:
@fernandomg could you please resolve conflicts?

fvictorio:
@vbaranov I tried to add a horizontal scroll but it turned out to be way harder than I thought, because the table is made with flexbox so adding a scrollbar is not as straightforward as doing `overflow-x: auto`.

I'd say we merge this and open a new issue for that bug, so that this feature is merged for today's demo. What do you think?

vbaranov:
Ok, let's go this way.

PR #505, opened 2018-01-18
(Refactor) Refactor RadioInputField and add tests

### LINKED ISSUE #442: (Feature) Make RadioInputField more generic (right now it can only have two options)

All

No

No

No

---

fvictorio:
Closes #442.

This PR makes RadioInputField acccept more than two options, and improves a little its code. It also adds tests for the component.

I thought about using it for the gas price selector, but that component is special because it shows an input when you select a custom gas price, so I ended up leaving it as it was.

### DIFF
--- a/src/components/stepThree/index.js
+++ b/src/components/stepThree/index.js
@@ -321,14 +321,10 @@ export class stepThree extends React.Component {
             description={`Minimum amount tokens to buy. Not a minimal size of a transaction. If minCap is 1 and user bought 1 token in a previous transaction and buying 0.1 token it will allow him to buy.`}
           />
           <RadioInputField
-            side="right"
+            extraClassName="right"
             title={ENABLE_WHITELISTING}
-            items={["yes", "no"]}
-            vals={["yes", "no"]}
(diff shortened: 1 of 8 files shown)


src/components/stepThree/index.js
```diff
@@ -364,7 +360,7 @@ export class stepThree extends React.Component {
             <div className="hidden">
               <div className="input-block-container">
                 <InputField
-                  side="left"
+                  extraClassName="left"
```
fernandomg: it should be `side` for the `InputField`

fvictorio: Oops, thanks. Fixed.

fvictorio:
You mean how to run tests for a single component? You have to pass the name of the component to `jest`:

```
npm run test:dapp -- ComponentName
```

(Actually, I think that runs all the tests that match `ComponentName`, but it's good enough for most scenarios)

If that's what you meant, I'll add it to the wiki.

vbaranov:
Yes, this is actually what I mean. It would be great if it would be documented in wiki.

PR #514, opened 2018-01-23
(Fix) Avoid duplicates in reserved tokens list

### LINKED ISSUE #441: (Feature) When reserving tokens, if the same address is input twice, replace the previous occurrence in the list

All

No

No

No

---

fvictorio:
Closes #441.

If address already exists in the list of reserved tokens, it's replaced with the new value.

Should we do the same for the whitelist component?


fvictorio:
Oh, that was intentional. I thought only one of the two options were allowed. I'll change that.

fvictorio:
Fixed.

fvictorio:
@vbaranov What bug do you mean? The behavior in the gif you shared is the expected one if I understood correctly.

vbaranov:
@fvictorio I expected 7% and 1 token, as a final distribution. No?

fvictorio:
Ohh, I see now. Thanks for clarifying.

vbaranov:
Sure, anytime

fvictorio:
@vbaranov fixed. Luckily, this also fixes #523.

PR #537, opened 2018-01-30
(Fix) Updated contracts after security audit

vbaranov:
As a result of security audit, contracts were updated with these PRs: https://github.com/poanetwork/ico/pull/14, https://github.com/poanetwork/ico/pull/15, https://github.com/poanetwork/ico/pull/16, https://github.com/poanetwork/ico/pull/17, https://github.com/poanetwork/ico/pull/18, https://github.com/poanetwork/ico/pull/19, https://github.com/poanetwork/ico/pull/20, https://github.com/poanetwork/ico/pull/21, https://github.com/poanetwork/ico/pull/22

Changes should be implemented in ICO Wizard due to audit:
- Finalization is in two steps: distribute reserved tokens (with batches) -> finalization

screen shot 2018-01-30 at 22 41 19
- `setTier` instead of `setLastCrowdsale` for PricingStrategy contract
- `setEarlyParticipantsWhitelist` renamed to `setEarlyParticipantWhitelistMultiple`
- `updateRate` is from crowdsale contract now (not from Pricing strategy). All properties of crowdsale are managed from crowdsale contract
- Pricing strategy has only one parameter in constructor

Also, there are some new restrictions for the crowdsale owner,
- can set reserved tokens list only once
- can join tiers only once
- can't change rate of the tier, when it is active or finished
- can't change supply of the tier, when it is active or finished
- can't change start time of the tier, when it is active or finished
- can't change end time of the tier, when it is finished

@fernandomg @fvictorio please let me know if any your tests will fail. After merging of this PR we could consider opening wizard for mainnet.

### DIFF
--- a/src/components/manage/index.js
+++ b/src/components/manage/index.js
@@ -208,17 +208,18 @@ export class Manage extends Component {
               .then(reservedTokensDestinationsLen => {
 
                 let batchesLen
-                let batchesInt = Number(reservedTokensDestinationsLen / addressesPerBatch)
-                if (reservedTokensDestinationsLen % addressesPerBatch > 0)
+                let batchesInt = parseInt(reservedTokensDestinationsLen / addressesPerBatch, 10)
+                if (reservedTokensDestinationsLen % addressesPerBatch > 0) {
                   batchesLen = batchesInt + 1
-                else
+                } else {
                   batchesLen = batchesInt
+                }
 
                 const distributeMethod = crowdsaleContract.methods.distributeReservedTokens(addressesPerBatch)
                 let opts = {
                   gasPrice: this.props.generalStore.gasPrice
                 }
-                let batches = new Array(batchesLen);
+                let batches = Array.from(Array(batchesLen).keys());
                 this.distributeReservedTokensRecursive(batches, distributeMethod, opts)
                 .then(() => {
                   successfulDistributeAlert()
@@ -439,7 +440,7 @@ export class Manage extends Component {
           <p className="title">Distribute reserved tokens</p>
           <p className="description">Reserved tokens distribution is the last step of the crowdsale before finalization.
             You can make it only after the end of the last tier. If you reserved more then 100 addresses for your crowdsale, the distribution will be executed in batches with 100 reserved addresses per batch. Amount of batches is equal to amount of transactions</p>
-          <Link to='#' onClick={() => this.distributeReservedTokens(1)}>
+          <Link to='#' onClick={() => this.distributeReservedTokens(100)}>
             <span className={`button button_${!canDistribute ? 'disabled' : 'fill'}`}>Distribute tokens</span>
           </Link>
         </div>
--- a/src/components/manage/index.js
+++ b/src/components/manage/index.js
@@ -203,17 +203,11 @@ export class Manage extends Component {
           crowdsaleContract.methods.token().call()
           .then(token => {
             attachToContract(contractStore.token.abi, token)
-            .then(tokenContract => { // eslint-disable-line no-loop-func
+            .then(tokenContract => {
               tokenContract.methods.reservedTokensDestinationsLen().call()
               .then(reservedTokensDestinationsLen => {
 
-                let batchesLen
-                let batchesInt = parseInt(reservedTokensDestinationsLen / addressesPerBatch, 10)
-                if (reservedTokensDestinationsLen % addressesPerBatch > 0) {
-                  batchesLen = batchesInt + 1
-                } else {
-                  batchesLen = batchesInt
-                }
+                const batchesLen = Math.ceil(reservedTokensDestinationsLen / addressesPerBatch)
 
                 const distributeMethod = crowdsaleContract.methods.distributeReservedTokens(addressesPerBatch)
                 let opts = {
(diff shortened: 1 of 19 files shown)


igorbarinov:
@fernandomg @fvictorio this is a task with priority. Please test it

fernandomg:
I have a question regarding the expected behavior with the whitelists.

Lets say I have a crowdsale with two tiers, and whitelist enabled. While creating the crowdsale, I added the addr `0x12345` (with MIN: 1, MAX: 10) to the 2nd tier.

In the manage screen, if I edit the first tier and add the addr `0x12345` to it (with MIN: 5, MAX: 20), and save. Both tiers, will have `0x12345` with MIN: 5, and MAX: 20.

Is it the expected behavior?

vbaranov:
Yes, if we keep in mind, that we use inheritance of whitelists strategy, it is expected behavior for me.

fvictorio:
Is the "Distribute tokens" button working? It just shows me a success modal but no transaction happens. And if I refresh, it's enabled again.

vbaranov:
It should work. Which network? Please try with Metamask, if you are using ganache.

fvictorio:
I tried both locally and in Rinkeby. I think the same happened to @fernandomg.

This is what I did:

- Created a crowdsale with reserved tokens for another address
- Bought all the supply
- Went to the manage page after the end date and clicked on "Distribute tokens"

vbaranov:
Ok, thanks, let me check. If you'll find something else, please let me know.

vbaranov:
It was an empty array of batches. Fixed. I will wait for your feedback again @fernandomg @fvictorio

fvictorio:
Yes, it works now.

There are some UI issues, but I don't think they should block the merge of this PR:

- If, before the crowdsale ends, you buy all the tokens and then in the manage page you click the "Distribute tokens" button, the "Finalize" button is enabled, but it can't really be used.
- The rate and supply inputs are not disabled, but if you modify them the transaction fails.
- Likewise, if a modifiable tier is active, its start date input is enabled, but you can't really modify it.

Again, I would create new issues for all those things and would work on them in separate PRs.

I will check the code now, but the functionality seems fine.

src/components/stepFour/index.js
```diff
@@ -513,7 +512,7 @@ const { PUBLISH } = NAVIGATION_STEPS
     const tokenAddr = token.addr
     const crowdsaleAddr = crowdsale.addr
 
-    setLastCrowdsaleRecursive(pricingStrategyABI, pricingStrategy.addr, crowdsaleAddr.slice(-1)[0])
+    setTierRecursive(pricingStrategyABI, pricingStrategy.addr, crowdsaleAddr)
```
fvictorio: I'm curious about this change, why were we passing the last tier only and now we are passing all the tiers?

src/components/manage/index.js
```diff
@@ -105,36 +108,146 @@ export class Manage extends Component {
...
+              tokenContract.methods.reservedTokensDestinationsLen().call()
+              .then(reservedTokensDestinationsLen => {
+
+                let batchesLen
+                let batchesInt = parseInt(reservedTokensDestinationsLen / addressesPerBatch, 10)
+                if (reservedTokensDestinationsLen % addressesPerBatch > 0) {
+                  batchesLen = batchesInt + 1
+                } else {
+                  batchesLen = batchesInt
+                }
```
fvictorio: I think this whole block can be replaced with:

```
let batchesLen = Math.ceil(reservedTokensDestinationsLen / addressesPerBatch)

File: src/components/manage/index.js [hunk 3]
```diff
@@ -105,36 +108,146 @@ export class Manage extends Component {
...
+        const { crowdsaleStore, contractStore } = this.props
+
+        if (!crowdsaleStore.selected.distributed && this.state.canDistribute) {
+          this.showLoader()
+
+          const crowdsaleContract = this.state.lastContract
+          crowdsaleContract.methods.token().call()
+          .then(token => {
+            attachToContract(contractStore.token.abi, token)
+            .then(tokenContract => { // eslint-disable-line no-loop-func
```
fvictorio: This eslint-disable is not necessary here

src/components/manage/index.js
vbaranov: Oh, sure, Franco. Thanks.

src/components/manage/index.js
vbaranov: Yes, fixed.

src/components/stepFour/index.js
vbaranov: We were storing `lastCrowdsale` property for `updateRate` method to check whether crowdsale is finalized or not until the audit. I moved updateRate method to crowdsale contract to have only one endpoint to change crowdsale data. Since that, we don't need to store `lastCrowdsale` property, but we need to join crowdsale with corresponding pricing strategy. This is why we need to store tier address in pricing strategy contract.

vbaranov:
Thanks @fvictorio I've created related issues:

https://github.com/poanetwork/ico-wizard/issues/541
>The rate and supply inputs are not disabled, but if you modify them the transaction fails.

https://github.com/poanetwork/ico-wizard/issues/542
>Likewise, if a modifiable tier is active, its start date input is enabled, but you can't really modify it.

https://github.com/poanetwork/ico-wizard/issues/543

PR #538, opened 2018-01-30
(Feature) Resume contract deployment

fernandomg:
This is the second step towards the completion of #464 

For `development` env modal with deployment steps will display a button to start the deployment, in production there will be no change in UX.

resume


fernandomg:
@vbaranov The deploy/resume buttons will not be visible in production. This PR was a refactor of the whole step four, so we can implement resume properly in the future without altering the current workflow.

vbaranov:
@fernandomg thanks for the clarification. It seems I missed your comment, that it won't affect production.

PR #548, opened 2018-02-01
(Fix) web3 build fix

vbaranov:
Closes https://github.com/poanetwork/ico-wizard/issues/534

**Problem**: web3 package is updated with merging upstream, but new build excluding handling of this error "Error: Transaction was not mined within 50 blocks, please make sure your transaction was properly send. Be aware that it might still be mined!" is not created
**Solution**: new build is created

Also:
- `oracles-combine-solidity` is renamed to `solidity-flattener`
- all `oraclesorg` dependencies are renamed
- `finalize` additional error handling on gas estimation


vbaranov:
@fernandomg @fvictorio please take a look

Febrero:
Still have issues #532 

Hello

the issue hasn't been fixed. just tried 3 times every time it fails.

1. i have ether in both addresses im using for crowdfunding (owner address, escrow address)
2. it seems the errors are always the same 2.
logfiles below.

[wizard.poa.network-1517582162205.log](https://github.com/poanetwork/token-wizard/files/1689702/wizard.poa.network-1517582162205.log)
[wizard.poa.network-1517582171355.log](https://github.com/poanetwork/token-wizard/files/1689703/wizard.poa.network-1517582171355.log)

fernandomg:
@Febrero indeed, I have just tested it with https://wizard.poa.network/ and it failed.

@vbaranov changes were properly deployed?

fvictorio:
Yes, according to Netlify, the changes were deployed.

vbaranov:
Actually, the problem exists with Netlify. It generates "old" .js file https://wizard.poa.network/static/js/main.a7fb81e5.js. You can find error "Transaction was not mined within 50 blocks, please make sure your transaction was properly send. Be aware that it might still be mined!".

But if you create local deployment and build with `npm run build`, the correct js is generated, where the handling of this error is commented.

vbaranov:
@fernandomg @fvictorio if you'll have any ideas, please let me know.

vbaranov:
For now, I have no idea except remove `--no-save` here:
 >cd submodules/poa-web3-1.0 && npm install && cd ../../ && npm install --no-save

We introduced it here: https://github.com/poanetwork/token-wizard/pull/512

vbaranov:
It doesn't work too. I checked in pre-deploy

fvictorio:
I'm looking into this. I triggered a deploy with a clean cache in Netlify to see if that fixes it.

My main suspect right now is this:

```
git submodule update --init --recursive --remote
```

The `--remote` flag is what troubles me: it will update the submodule to the last commit in the pointed flag. That means that it's not going to use the submodule that is commited.

That being said, when I built the app after updating the submodules both with and without the `--remote` flag, the result was the same.

I'll wait for the build to finish to see if there was some change.

fvictorio:
Ok, the problem seems to be with Netlify using npm 3. I created a PR with a `.nvmrc` file so that Netlify uses node 8.9.1 and npm 5. I'll check the deploy preview to see if this effectively fixes the issue.

PR #556, opened 2018-02-02
(Fix) Allow finalize after distribute tokens

### LINKED ISSUE #541: (Fix) crowdsale owner should be able to finalize, if all tokens are sold before crowdsale ending

fernandomg:
Closes #541


vbaranov:
@fernandomg @fvictorio it should be reverted. I think it won't work. This is why:
1. `isMinimumGoalReached` every time is true because the token wizard doesn't use soft cap. It is always zero. `isCrowdsaleFull` property instead checks that crowdsale is reached hard cap.
2. When we get properties like `finalized`, `areReservedTokensDistributed ` we should call them from the last tier. So we need to call `getLastTier` in the current contract and then call all these properties from the last tier contract.

Issue #529, opened 2018-01-28
(Feature) Implement local storage for mobx state


igorbarinov:
Problem:
when a user refreshes a page he loses the state of the crowdsale and not able to download the source code. Implementation of the feature allows users to continue set up of a crowdsale in the same browser after refresh

Solution:
store mobx state in local storage state after each interaction

---

Pending:

- [x] Persist stores in local storage
- [x] Clear storage properly after the deploy is finished
- [x] Save `deploymentStep` after each step (not just when a transaction fails)
- [x] Notify that there is an incomplete deploy if user opens the home page of the wizard
- [x] Don't show the progress modal if user refreshes the page after all steps are done
- [ ] Show skip button if transaction fails

fvictorio:
We discussed a little about this with @fernandomg and while persisting the mobx state in local storage is a part of the solution, there are several things that need to be taken into account (when is the storage cleaned, what happens if the user's browser crashes and they open the wizard from the beginning, how the persistence interacts with non-serializable objects like web3).

While we work on all that stuff, a simple thing we can implement is an alert when the user refreshes or closes the tab in the middle of the deployment, to reduce the chances of someone accidentally losing ether.

@igorbarinov I'll create an issue for that if you think it's a good idea.

PR #564, opened 2018-02-05
(Feature) Create component to prevent page refresh

fernandomg:
This is an intermediate solution before #529.

Alerts the user if he/she is about to leave or refresh the current page (this is active in step four only)

prevent_refresh

### DIFF
--- a/src/components/Common/PreventRefresh.js
+++ b/src/components/Common/PreventRefresh.js
@@ -1,0 +1,19 @@
+import React, {Component} from 'react'
+
+export class PreventRefresh extends Component {
+  onUnload = (e) => {
+    e.returnValue = "You'll loose current deployment, are you sure?"
+  }
+
+  componentDidMount () {
+    window.addEventListener('beforeunload', this.onUnload)
+  }
+
+  componentWillUnmount () {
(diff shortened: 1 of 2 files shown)


src/components/Common/PreventRefresh.js
```diff
@@ -0,0 +1,19 @@
+import React, {Component} from 'react'
+
+export class PreventRefresh extends Component {
+  onUnload = (e) => {
+    e.returnValue = "You'll loose current deployment, are you sure?"
```
fvictorio: This message is not the one that is being shown.

Besides, I think it should be `You'll lose your current deployment, are you sure?`

fernandomg: Typo fixed.

For the alert to be displayed it's required to set some value for `returnValue`, despite it's not being used by browsers: 
1. https://www.chromestatus.com/feature/5349061406228480 
2. https://developer.mozilla.org/en-US/docs/Web/Events/beforeunload

vbaranov:
Please do not merge this. Because someone has started crowdsale creation on mainnet. I'll merge it later.

PR #565, opened 2018-02-05
(Fix) method to verify if crowdsale can be finalized

### LINKED ISSUE #541: (Fix) crowdsale owner should be able to finalize, if all tokens are sold before crowdsale ending

fernandomg:
Closes #541 

With fixes pointed in: https://github.com/poanetwork/token-wizard/pull/556#issuecomment-363105877

> 1. isMinimumGoalReached every time is true because the token wizard doesn't use soft cap. It is always zero. isCrowdsaleFull property instead checks that crowdsale is reached hard cap.
> 2. When we get properties like finalized, areReservedTokensDistributed we should call them from the last tier. So we need to call getLastTier in the current contract and then call all these properties from the last tier contract.

### DIFF
--- a/src/components/manage/index.js
+++ b/src/components/manage/index.js
@@ -161,61 +161,23 @@ export class Manage extends Component {
     })
   }
 
-
   canFinalize = () => {
     const { contractStore } = this.props
+    const lastCrowdsaleAddress = contractStore.crowdsale.addr.slice(-1)[0]
 
-    return new Promise((resolve, reject) => {
-      if ((!this.state.crowdsaleHasEnded) || (this.state.shouldDistribute && this.state.canDistribute)) {
-        this.setState({ canFinalize: false })
-        resolve(this.state.canFinalize)
+    return attachToContract(contractStore.crowdsale.abi, lastCrowdsaleAddress)
+      .then(crowdsaleContract => crowdsaleContract.methods.isCrowdsaleFull().call())
+      .then(
+        (isCrowdsaleFull) => {
+          const { crowdsaleHasEnded, shouldDistribute, canDistribute } = this.state
+          const wasDistributed = shouldDistribute && !canDistribute
 
-      } else {
-        const lastCrowdsaleAddress = contractStore.crowdsale.addr.slice(-1)[0]
-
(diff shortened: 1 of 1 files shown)


src/components/manage/index.js
```diff
@@ -163,29 +163,56 @@ export class Manage extends Component {
...
+                  ])
+                    .then(([finalized, isCrowdsaleFull, areTokensDistributed, reservedTokensLen]) => {
+                      let _canFinalize
+
+                      if (isCrowdsaleFull && !finalized) {
+                        // If all the tokens were sold and the crowdsale is not already finalized,
+                        // it can be finalized if there are no tokens to distribute, or if they were already distributed
+                        _canFinalize = parseInt(reservedTokensLen, 10) === 0 || areTokensDistributed
+                      } else {
+                        _canFinalize = false
```
vbaranov: Let's imagine that crowdsale is not finalized, ended, but hard cap is not sold, reserved tokens are distributed. Why can we not to finalize?

fvictorio: The `finalize()` method in the crowdsale contract has na `inState(State.Success)` modifier, and, if I'm understanding this correctly, for the crowdsale to be in `State.Success` the minimum goal has to have been reached. Am I missing something?

vbaranov: Right, for crowdsale in `State.Success` the minimum goal is reached. But if the minimum goal is reached, it doesn't mean that crowdsale is in `State.success` state.

fvictorio: Oh, ok. Should we just check that `getState() === 4` then?

vbaranov: I think the full logic schematically should be rewritten as:
- if (isCrowdsaleFull && ((shouldDistribute && !canDistribute) || !shouldDistribute) => canFinalize = true
- else if (crowdsaleHasEnded && ((shouldDistribute && !canDistribute) || !shouldDistribute) => canFinalize =true
- else canFinalize = false

fernandomg: @vbaranov what you think about it now? I'd tested it with different scenarios and worked fine.

PR #567, opened 2018-02-06
(Fix) loader not loading after deployment has finished

### LINKED ISSUE #566: (Fix) all steps are finished, but loader is not hidden

Mainnet

No

photo5172861386307512277

No

---

fernandomg:
Closes #566 

Seems that the user closed the deployment steps modal, and the process was lacking of the option to close the _loading_ overlay.

### DIFF
--- a/src/components/stepFour/index.js
+++ b/src/components/stepFour/index.js
@@ -23,7 +23,6 @@ import { toast } from '../../utils/utils'
 import { StepNavigation } from '../Common/StepNavigation'
 import { DisplayField } from '../Common/DisplayField'
 import { DisplayTextArea } from '../Common/DisplayTextArea'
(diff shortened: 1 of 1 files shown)


fvictorio:
Should we also change the `userHideModal` function? The second line:

```
    if (!this.props.deploymentStore.deploymentHasFinished) this.setState({ loading: true })
```

seems redundant now, doesn't it?

fernandomg:
@fvictorio you mean to change it to:
```
this.setState({ loading: !this.props.deploymentStore.deploymentHasFinished })
```
?

fvictorio:
I'd just remove it, and handle the loader independently of the modal.

fernandomg:
I did removed the `Loader

PR #574, opened 2018-02-07
(Feature) Header and footer on iframe

### LINKED ISSUE #530: (Feature) Remove footer and header in embedded mode
Problem:
When wizard invest page embedded in iframe there is header/footer of ICO Wizard inside the iframe. Users of ICO Wizard decide not embed the page because it's not convenient/not design friendly.
Example of an embedded invest page https://www.thebtcbtq.com/btq
https://cl.ly/143d3c0U2U1O

Solution:
When invest page is in an iframe, don't show footer and header. Should be configurable in a yml config file. By default header/footer should be disabled in the embedded mode.

fernandomg:
Closes #530 

The `showHeaderAndFooterInIframe` attribute was added to the ex `ICOConfig` variable (now `CrowdsaleConfig`)

Even removing the header, if the iframe's height is `600px`, the QR code won't be visible in the invest page.  As a consequence Parts of the page where hidden when the window is `600px` or less in height, to make the important information visibile.

image

After approval/merge of this PR, I'll update the wiki, as requested: https://github.com/poanetwork/token-wizard/issues/530#issuecomment-363785213


src/components/invest/index.js
```diff
@@ -375,7 +375,7 @@ export class Invest extends React.Component {
             <a href="https://kovan.etherscan.io/tx/0x42073576a160206e61b4d9b70b436359b8d220f8b88c7c272c77023513c62c3d">transaction</a> on Kovan network.
           </p>
         </div>
-        <div className="invest-table-cell invest-table-cell_right">
+        <div className={`invest-table-cell invest-table-cell_right${investThrough === INVESTMENT_OPTIONS.QR ? ' qr-selected': ''}`}>
```
fvictorio: We should consider adding https://github.com/JedWatson/classnames

fernandomg: Thanks for the suggestion, just added the utility

PR #575, opened 2018-02-07
(Feature) Alert if not the owner and prevent editing

### LINKED ISSUE #552: (Feature) Manage page: no alert if I am not the owner of crowdsale
**Problem**: Manage page is accessible for reading for everyone. And buttons are enabled if I am not the owner of crowdsale. Of course, managing transactions will fail. 

**Solution**: But maybe let's disable all buttons if the current user is not the creator of the crowdsale and add alert for manage page, that current user is not the owner of the crowdsale and cannot manage it.

fernandomg:
Closes #552

### DIFF
--- a/src/components/manage/index.js
+++ b/src/components/manage/index.js
@@ -6,11 +6,18 @@ import { CONTRACT_TYPES, TEXT_FIELDS, TOAST, VALIDATION_MESSAGES } from '../../u
 import { InputField } from '../Common/InputField'
 import '../../assets/stylesheets/application.css'
 import { WhitelistInputBlock } from '../Common/WhitelistInputBlock'
-import { successfulFinalizeAlert, successfulDistributeAlert, successfulUpdateCrowdsaleAlert, warningOnFinalizeCrowdsale } from '../../utils/alerts'
+import {
+  successfulFinalizeAlert,
+  successfulDistributeAlert,
+  successfulUpdateCrowdsaleAlert,
+  warningOnFinalizeCrowdsale,
+  notTheOwner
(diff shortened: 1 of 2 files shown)


src/components/manage/index.js
```diff
@@ -6,11 +6,18 @@ import { CONTRACT_TYPES, TEXT_FIELDS, TOAST, VALIDATION_MESSAGES } from '../../u
...
+  successfulDistributeAlert,
+  successfulUpdateCrowdsaleAlert,
+  warningOnFinalizeCrowdsale,
+  notTheOwner
+} from '../../utils/alerts'
 import { getNetworkVersion, sendTXToContract, attachToContract, calculateGasLimit } from '../../utils/blockchainHelpers'
 import { getWhiteListWithCapCrowdsaleAssets, toast } from '../../utils/utils'
 import { contractsInfo, getTiers, processTier, updateTierAttribute } from './utils'
 import { Loader } from '../Common/Loader'
+import web3Store from '../../stores/Web3Store'
```
vbaranov: Do we need it here? We injected `web3Store` below.

fvictorio: Unrelated to this PR, but we could use something like this https://github.com/coopermaruyama/react-web3 instead of a store, since stores should have plain data, not a complex object like web3. Besides, using a component like that, we'd be sure that web3 is present, available and unlocked when the component is rendered.

fernandomg: @vbaranov was imported by the IDE, just deleted it. Thanks for the catch!

PR #576, opened 2018-02-07
(Feature) Autosave all stores and clear localStorage after stepFour

fvictorio:
This is the first step in the local storage feature. Please note that this PR is using a feature branch (`persist-crowdsale-state-#529`) as the base branch, since this is not ready for production yet. I expect this feature to have a lot of corner cases, so doing multiple reviews and PRs should help us to eventually merge this in master with less risk.

In this change, all the stores are autosaved to localStorage, and get their data from there in the constructor. This makes the deploy to be resumable: to test it, just reject a transaction and refresh the page; it should allow you to continue the deploy. (As a side effect, that might be a feature or a bug, you can also refresh in the step two and three and most inputs are preserved.)

**The main issue** with this is that if you deploy a crowdsale and then go back to the start and try to deploy a new one, some of the data from the previous crowdsale is already loaded. If you want to try several deploys, you have to go to the home page, open devtools, run `localStorage.clear()` and refresh the page.

(The same issue happens if you just visit a crowdsale or invest page and then try to deploy your own crowdsale: the data from the crowdsale you visited is pre-loaded.)

**Another major issue** is that for you to resume the deploy, you have to go to step four. So, for example, if your browser crashes in the middle of a deploy, and you go to the main page of the wizard, you'll not be redirected to the deploy.

My idea for improving this is that if you go to the home page of the wizard and the data in local storage indicates that you have an incomplete deploy, a modal appears telling you this and asking if you want to resume the deploy. If you accept, you are redirected to the step four and things should just work. If you cancel, the local storage data is cleared.


src/stores/GasPriceStore.js
```diff
@@ -74,6 +64,31 @@ class GasPriceStore {
         return Promise.resolve()
       })
   }
+
+  @computed
```
fvictorio: I converted the descriptions to computed values because functions cannot be serialized and saved in local storage.

vbaranov:
>My idea for improving this is that if you go to the home page of the wizard and the data in local storage indicates that you have an incomplete deploy, a modal appears telling you this and asking if you want to resume the deploy. If you accept, you are redirected to the step four and things should just work. If you cancel, the local storage data is cleared.

I agree. We need this to implement. Also, we have a similar bug in the current version: when we successfully passed through the deployment process and returned back to the main page again, for example, through the left-top logo, all fields in step 2,3 are filled from the previous deployment. I think it should be fixed in the frame of future PR implementing local storage too.

Issue #542, opened 2018-02-01
(Fix) Manage page: the rate and supply inputs are not disabled, if crowdsale is started


vbaranov:
https://github.com/poanetwork/ico-wizard/pull/537#issuecomment-362088690

>The rate and supply inputs are not disabled, but if you modify them the transaction fails.

hashguide:
I know before we were able to change the rate and supply of the crowdsale after it has started as well, just the start time was unable to. Has this change due to the security audits or is this the intended way? 

Allowing to change the rate while crowdsale is running allows you to use single tier crowdsale to have a multi-tier bonus system of some sort without having to deploy the same amount of tiers as rate changes.

vbaranov:
Yes, it was one of the points indicated in security audit. We closed this possibility due to it. For now, owner should create multiple tiers to implement flexible bonus system.

hashguide:
Ok sounds good. Just wanted to be clear about it as well.

PR #583, opened 2018-02-08
(Fix) Disable date, rate and supply fields if crowdsale has started

### LINKED ISSUE #543: (Fix) Start date input is not disabled, if crowdsale is started

>Likewise, if a modifiable tier is active, its start date input is enabled, but you can't really modify it.

fernandomg:
Closes #543 & #542 

Allow only to add addresses to the tier's whitelist if the tier has started already.


fvictorio:
This lets you edit the rate input when the tier has not started, but the transaction fails. Maybe the contract doesn't allow changing the rate at all. The start date, end date and supply can be modified before the tier starts.

@vbaranov Is that the expected behavior? If so, we should always have the rate input disabled.

src/components/manage/index.js
```diff
@@ -519,7 +524,7 @@ export class Manage extends Component {
     }
 
     const tierStartAndEndTime = (tier, index) => {
-      const disabled = !this.state.ownerCurrentUser || !tier.updatable || crowdsaleHasEnded
+      const disabled = !this.state.ownerCurrentUser || !tier.updatable || this.tierHasStarted(index) || crowdsaleHasEnded
```
vbaranov: End time should be editable if crowdsale is started and not finished yet (if tier is editable)

vbaranov:
>This lets you edit the rate input when the tier has not started, but the transaction fails. Maybe the contract doesn't allow changing the rate at all. The start date, end date and supply can be modified before the tier starts.

If the tier is editable and not started yet, a transaction to change rate shouldn't fail. Do you have another behavior @fvictorio ?

fvictorio:
Yes, the transaction fails.

vbaranov:
@fvictorio I have successfully changed the rate for crowdsale, that is not started yet.
These are conditions to be able to change rate:
```
function updateRate(uint newOneTokenInWei) public onlyOwner {
    assert(!finalized);
    assert(isUpdatable);
    assert(now <= startsAt);
    ...
    assert(!lastTierCntrct.finalized());
    ...
  }
```
Are all of them satisfied for your try?

vbaranov:
@fvictorio I have successfully changed the rate for crowdsale, that is not started yet.
These are conditions to be able to change rate:
```
function updateRate(uint newOneTokenInWei) public onlyOwner {
    assert(!finalized);
    assert(isUpdatable);
    assert(now <= startsAt);
    ...
    assert(!lastTierCntrct.finalized());
    ...
  }
```
Are all of them satisfied for your try?

fvictorio:
The problem was with the `estimateGas` function and ganache. The same thing had happened before in the invest page. It's not related with this PR.

src/components/manage/index.js
fernandomg: Mh! I'll need #586 to be merged before fixing this, as fixes in that PR are required for this change.

fernandomg: change done

PR #588, opened 2018-02-09
Check if there is an unfinished deploy

fvictorio:
Part of #529.

Show a message if there is an incomplete deploy, and allow resuming or canceling it:

image

The styles could be better, and the "Cancel" button should show a warning explaining that the user will lose all the gas inverted in the deploy. But I want to check if this approach makes sense to you before doing further changes.

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,6 +1,8 @@
 import React, { Component } from 'react';
+import { inject, observer } from 'mobx-react'
 import './assets/stylesheets/application.css';
 import { Header, Footer, Home, Manage, stepOne, stepTwo, stepThree, stepFour, Crowdsale, Invest } from './components/index'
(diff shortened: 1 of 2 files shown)


fvictorio:
Please don't merge this yet so that we can use the deploy preview in the demo.

fvictorio:
I agree with the idea, but I'm concerned about how allowing to see the manage, crowdsale and invest pages can affect the stores. If you go to a different crowdsale, the address of the crowdsale is going to be modified and your persisted state is going to be left in an inconsistent state.

The best solution in my opinion is not persist (nor load from the localStorage) the stores in those pages, but I'm not sure how hard will that be.

vbaranov:
Ok, I see. Let's leave it as is

PR #598, opened 2018-02-15
(Feature) Persist crowdsale state

### LINKED ISSUE #529: (Feature) Implement local storage for mobx state

fvictorio:
Closes #529.

This is the final PR for the local storage feature. The code was already reviewed in the PRs done against the feature branch, so we should focus our efforts in testing all possible scenarios.


fvictorio:
I agree with the concerns in both points, but the solutions can be a little hard to do right now.

With respect to the skip button, I'm not sure if disabling it selectively is the solution: there will always be corner cases that we didn't consider. Besides, doing that would mean changing a lot the way the deployment is done now (it can be done, of course, but it will mean postponing this feature even more). I think a temporary solution would be adding a warning indicating the user to only skip it if they are completely sure that that transaction was already done.

Regarding the transaction hash, I think that is by far the best solution, but this too would mean a lot of changes. If I remember correctly, what we discussed in the last demo was that the skip button was the quick (though not ideal) approach that we'd do for now. If improving that is a priority, we can start designing that change after merging this.

vbaranov:
>I think a temporary solution would be adding a warning indicating the user to only skip it if they are completely sure that that transaction was already done.

Yes, some kind of warning alert on *Skip transaction* button is a good idea.

>Regarding the transaction hash, I think that is by far the best solution, but this too would mean a lot of changes. If I remember correctly, what we discussed in the last demo was that the skip button was the quick (though not ideal) approach that we'd do for now. If improving that is a priority, we can start designing that change after merging this.

I have added to the list of tasks https://github.com/poanetwork/token-wizard/issues/601, whatever priority will be agreed.

fvictorio:
@vbaranov I added the warning on Skip transaction.

fvictorio:
Since this is a feature with high risk, maybe it would be a good idea to merge it on Monday. Are you OK with that, @vbaranov?

vbaranov:
@fvictorio if we don't plan to change anything in this PR, it is better to merge it ASAP. @dennis00010011b if you will have a chance, please test this PR too - you have a great tool. It is deployed here: https://5a87189fefbe5d658ef89ee4--ico-wizard.netlify.com. Anyways, we can't merge without the 2nd approvement. And if we will be able to merge it before Monday I or @igorbarinov will revert it if we'll hear problems from users on weekend.

fvictorio:
Ok, sure. I think @fernandomg already did some testing around this, so if he and @acravenho approve it, we can merge it now.

dennis00010011b:
I got issue https://github.com/poanetwork/token-wizard/issues/602

fvictorio:
Oh, that is weird. But it's probably something to be worked on in a different PR.

dennis00010011b:
Issue https://github.com/poanetwork/token-wizard/issues/604

vbaranov:
@fernandomg, how does this PR look to you? @fvictorio, does anything stop us to merge it to master?

fvictorio:
No, I think it can be merged if @dennis00010011b and @fernandomg approve it.

dennis00010011b:
everything looks good, I can't find any issue

vbaranov:
Thank you, guys. I'll wait for a window between user's crowdsale creations to merge it.

Issue #603, opened 2018-02-17
(Bug) stuck tx status


lindsmcw:
If you are reporting a problem with Token Wizard, please include the following information:

### Which network did you use? (Mainnet, Kovan, Rinkeby, etc.)

Main network

### If you were able to create it, what is the URL of your crowdsale?

Not finished

### Do you have screenshots showing the problem?

Yes
![image](https://user-images.githubusercontent.com/9313857/36342916-fd59eee8-1450-11e8-8109-aece1d0dff4f.png)

### Do you see errors in the dev console? If yes, please include a screenshot
![image](https://user-images.githubusercontent.com/9313857/36342978-a0a6254e-1451-11e8-8a6c-72dfc8b2a4a6.png)
![image](https://user-images.githubusercontent.com/9313857/36342988-cc88b262-1451-11e8-9b42-c51c6d21a0fd.png)
![image](https://user-images.githubusercontent.com/9313857/36343024-37a73f96-1452-11e8-8df9-29fc6e71c018.png)
 at f.performUpdateIfNecessary (main.7c06d6f5.js:1)

To open the dev console in Google Chrome, press F12, or go to `View -> Developer -> Developer Tools`, and open the Console tab

If you see errors, please right click on them and "Save as..". Zip saved file and attach it to the Issue.

*Your answer*

---
The issue is there were 17 transactions required through the metamask network and it stopped after the 11th one. Metamask doesn't show the submit button for the next one.. how can I continue with this?

vbaranov:
The first error from the screenshot is ignored and it shouldn't block deployment process. You can see, that deployment continued after this error. Anyways, could you clarify some details for investigation:

1. Could you attach Zip archive of logs - it is hard to restore chronology from screenshot
2. Why do you say 17 transactions? If you have only 1 tier, you need to sign only 14 txs. I see, that you are trying to create crowdsale with 1 tier.
3. Did you do any manipulations with MetaMask to repeat some of the transactions?

lindsmcw:
1) Hi, I don't have that page open anymore, as it was stuck. Not able to do
anything, so I just saved this info then the window was closed:


main.7c06d6f5.js:1 stepThree ƒ i(){var t,e,n,r;E(this,i);for(var
o=arguments.length,a=Array(o),s=0;s<o;s++)a[s]=arguments[s];return
[...]

PR #605, opened 2018-02-19
(Fix) check isMined on error event

vbaranov:
We shouldn't handle an error from the previous deployment tx, if it is already mined in `repeatPolling()` of tx receipt. Possible fix for https://github.com/poanetwork/token-wizard/issues/603

### DIFF
--- a/src/utils/blockchainHelpers.js
+++ b/src/utils/blockchainHelpers.js
@@ -152,6 +152,7 @@ let sendTX = (method, type) => {
   return new Promise((resolve, reject) => {
     method
       .on('error', error => {
+        if (isMined) return
         console.error(error)
         // https://github.com/poanetwork/token-wizard/issues/472
         if (


vbaranov:
@fernandomg 
*Failed to check for transaction receipt* error is skipped by us [here](https://github.com/poanetwork/token-wizard/blob/master/src/utils/blockchainHelpers.js#L158) and [here](https://github.com/poanetwork/token-wizard/blob/master/src/utils/blockchainHelpers.js#L172). Thus, the problem in #603 is not directly caused by this error. We see that the deployment process was going further after this error. And the transaction status was got by polling of tx receipt. But I think that later, this transaction raised `error` event. And the deployment process stuck. But we have already got success status of this tx by polling of tx receipt before and we don't need to handle the error from this tx. This is my vision. Please, let me know if you have another one.

fernandomg:
@vbaranov
> Failed to check for transaction receipt error is skipped by us here and here.

Indeed, I was trying to make that point. And I assumed that #603 error was related to that.

But in the other hand, I don't understand how or why a `tx` will raise an `error` after returning an `status === 1`. I assumed that the `error` event will be triggered only when the `status === 0`.

vbaranov:
>I assumed that the error event will be triggered only when the status === 0.

no, actually transaction can be successfully mined, but `error` event is returned by any reason.
We have 3 events: `error`,` transactionHash`, `receipt`.
In `transactionHash` we are additionally polling tx receipt, in case of` receipt` event will not work. So, it is an additional insurance to get a receipt.
If `receipt` event is called, there is no problem - `error` event will never be called.
But if we have got a successful result from polling of receipt, that we started in `transactionHash` event, events `error` and `receipt` still can be thrown. And if later it will be `error` event, and it is not among those, that we are skipping, deployment will stuck.

PR #606, opened 2018-02-19
(Fix) Add validation to decimals

### LINKED ISSUE #568: (Fix) add validation to decimals field

Each network

No

No. Users are trying to insert float values to the field. Should be validated.

Uncaught (in promise) Error: Error: [number-to-bn] while converting number "0.001" to BN.js instance, error: invalid number value. Value must be an integer, hex string, BN or BigNumber instance. Note, decimals are not supported. Given value: "0.001"
   at l (utils.js:64)
   at Object.f [as toTwosComplement] (utils.js:77)
   at o.s [as _inputFormatter] (formatters.js:44)
   at o.encode (type.js:188)
   at index.js:255
   at Array.map (<anonymous>)
   at m.encodeParameters (index.js:254)
   at microservices.js:32
   at new Promise (<anonymous>)
   at o (microservices.js:14)
   at e.getEncodedABIClientSide (microservices.js:8)
   at utils.js:38
   at <anonymous>

---

fernandomg:
Closes #568 

Created a new component (`NumericInput`) whose intention is to be used across all the wizard where a number is required.

I just updated the **DECIMALS** input field for `stepTwo` with this new component

### DIFF
--- a/src/components/Common/NumericInput.js
+++ b/src/components/Common/NumericInput.js
@@ -1,0 +1,92 @@
+import React, { Component } from 'react'
+import { InputField } from './InputField'
+import { VALIDATION_TYPES } from '../../utils/constants'
+import update from 'immutability-helper'
+import { observer } from 'mobx-react'
+import { countDecimalPlaces } from '../../utils/utils'
+
+const { VALID, INVALID } = VALIDATION_TYPES
+
+@observer
+export class NumericInput extends Component {
+  constructor (props) {
+    super(props)
+
+    this.state = {
+      value: '',
+      validation: {
+        value: {
+          pristine: true,
+          valid: INVALID
+        }
+      }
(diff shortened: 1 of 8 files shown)


src/components/Common/NumericInput.js
```diff
@@ -0,0 +1,92 @@
...
+const { VALID, INVALID } = VALIDATION_TYPES
+
+@observer
+export class NumericInput extends Component {
+  constructor (props) {
+    super(props)
+
+    this.state = {
+      value: '',
+      validation: {
```
fvictorio: Since `NumericInput` can only have one value, I think this can be flattened to `this.state = { value, pristine, valid}`

fernandomg: done, thanks!

Issue #596, opened 2018-02-14
(Bug) Modifiable crowdsale: incorrect behavior if user tries to add existing whitelisted address


dennis00010011b:
### Steps to reproduce:
1. Create modifiable crowdsale with whitelisted addresses
2. Open manage 
3. Add in the whitelist existed whitelisted address, fill out min/max value
4. Click button "+"
5. Click button "Save"

Expected result: should be error message like "address already exist ..."
                           OR user should be allowed to modify existed addresses(Feature)
Actual result: button "Save' is enabled, user can proceed, message "Congrats. You've successfully updated the crowdsale "

If you are reporting a problem with Token Wizard, please include the following information:

### Which network did you use? (Mainnet, Kovan, Rinkeby, etc.)

*Rinkeby*

### If you were able to create it, what is the URL of your crowdsale?

*https://wizard.poa.network/invest?addr=0x011C0608e9858f22564C31199438f9a732B6f157&networkID=4*

### Do you have screenshots showing the problem?

![incorrbehavior](https://user-images.githubusercontent.com/14067096/36226895-1387be7c-1184-11e8-8601-06db3654e595.png)

https://drive.google.com/open?id=1iAliRoG85jqjbjE1Ws_B-Q33yJsUCUSy

### Do you see errors in the dev console? If yes, please include a screenshot

To open the dev console in Google Chrome, press F12, or go to `View -> Developer -> Developer Tools`, and open the Console tab

If you see errors, please right click on them and "Save as..". Zip saved file and attach it to the Issue.

*No*

---

fernandomg:
@dennis00010011b @vbaranov what you think about adding the address to the list, but highlighting it and with a warning icon?

Using the `title` attribute:
![image](https://user-images.githubusercontent.com/3315606/36741537-25d14202-1bc4-11e8-9837-868952e32f77.png)

Adding a component for tooltips (https://github.com/wwayne/react-tooltip):
![image](https://user-images.githubusercontent.com/3315606/36741607-4e91b190-1bc4-11e8-9f36-b4c2fc490f5b.png)

Please, let me know which approach you think is the best. Or if you have any other idea to implement this solution.

vbaranov:
@fernandomg personally, I like the 2nd one with react-tooltip. My worry is about how to revert this change before saving if a user forgot how many it was reserved initially (I guess, only with refreshing the page). Maybe, we can add a small revert button, also? It will get saved values for this address in smart-contracts.

[...]

PR #639, opened 2018-02-28
(Feature) Bulk import whitelist

### LINKED ISSUE #600: (Feature) Bulk import on whitelist addresses
Hi, 

I tried using POA Network for crowdsale demo test and really liked it. I will like to know how i can add bulk white-list addresses when creating a new contract? or if the feature doesn't exists any timeline when this feature will be availble?

Thanks,
Sagar

fvictorio:
Closes #600.

This is a very basic implementation of the whitelist bulk import. There are some things tan can be improved, but I wanted to make this PR as soon as possible so as to not generate merge conflicts with the work that @fernandomg is doing for #596.

Pending things:

1. Give some kind of feedback to the user. Something like: `300 addresses were added, 20 lines were invalid` (or something along those lines).
2. Better validation for addresses. Right now only the length is checked.
3. Better styling. Regarding this, it would be nice to add something like [Font Awesome](https://fontawesome.com/) for the icons. We could use it later to replace the images with icons.


pablofullana:
@fvictorio shall we open a new issue for items 1..3 that you mention in this PR description?

fvictorio:
Yes, I will create them.

PR #650, opened 2018-03-02
(Fix) Don't continue to step 4 if min cap is not valid

### LINKED ISSUE #621: (Fix) investor global min cap can be greater than total supply of the tiers
**Problem**: investor global min cap can be greater than total supply of the tiers
**Solution**: validate *investor min cap field* on *Continue* button, check if it is greater than total supply of all tiers.

fvictorio:
Closes #621.

This checks that the min cap isn't greater than _any_ supply, not the sum of all the supplies. Otherwise, it may be the case that you can't buy in any tier.

For example, if you have a tier with supply 10 and another with supply 5, and your min cap is 11, it's less than the sum of all supplies, but you can't buy in none of the tiers.

I don't like at all the error message in this fix: it shows a message that basically says "this field is wrong, and one of these three things is the one causing the error".

It would be much better UI to show only the messages that are relevant. For example, if you input "-1" it just says "Value must be positive". If you input "1.333" and the token only has two decimals, it says "Decimals should not exceed...". If you input "-1.333", it should show both messages.

I didn't do it that way in this PR because it's a much bigger change, but if you agree with this, I will create an issue to improve it.

### DIFF
--- a/src/components/stepThree/index.js
+++ b/src/components/stepThree/index.js
@@ -139,7 +139,7 @@ export class stepThree extends React.Component {
 
     const { tierStore, gasPriceStore } = this.props;
     const gasPriceIsValid = gasPriceStore.custom.id !== this.state.gasPriceSelected || this.state.validation.gasPrice.valid === VALID
-    const isMinCapValid = tierStore.globalMinCap <= tierStore.minSupply
+    const isMinCapValid = tierStore.globalMinCap <= tierStore.maxSupply
 
     for (let index = 0; index < tierStore.tiers.length; index++) {
       tierStore.validateTiers("endTime", index);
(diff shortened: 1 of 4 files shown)


fvictorio:
Yes, I did that intentionally. Otherwise, you won't be able to buy anything at all in the second tier. But if you think we should use the sum of all the supply instead of the minimum, I can change it.

vbaranov:
>Otherwise, you won't be able to buy anything at all in the second tier

Right, but I think this case shouldn't prevent crowdsale creation. Logic could be: If a user can't buy on any tier, we should prevent creation. But if any tier has supply more or equals a user's min cap, we shouldn't prevent.

fvictorio:
@vbaranov Updated. Now I check the global min cap against the max supply instead (if the max supply is less than the global min cap, then every tier has a supply that is less than the global min cap).

vbaranov:
@fvictorio This configuration doesn't work for me:

screencapture-deploy-preview-650-ico-wizard-netlify-3-1520352662677

fvictorio:
That's strange, I typed the same values and it worked for me. I tried both typing the tiers first and the min cap second, and the other way around.

vbaranov:
Yes, it is not repeated to me anymore too.

fvictorio:
Maybe it's an issue with a very specific sequence of steps (although I hope not). Let's keep our eyes open in case it happens again.

Issue #642, opened 2018-02-28
(Improvement) Improve styling of whitelist bulk import button


fvictorio:
See #639. Consider adding Font Awesome.

PR #656, opened 2018-03-05
(Feature) Indicate to user how many addresses were imported to the whitelist

### LINKED ISSUE #640: (Improvement) Give feedback to user when bulk importing whitelist addresses
See #639.

fvictorio:
Depends on #642. Closes #640.

Give some very basic feedback to the user after importing a whitelist CSV. We can iterate over this to give better feedback if needed (for example, this doesn't show how many malformed rows were skipped, nor indicates if the imported CSV had addresses that were already added, and therefore ignored).


vbaranov:
@fernandomg if you have a chance, please, review. We need the 2nd signature because this is the feature.

Issue #578, opened 2018-02-08
(Fix) add validation for Whitelist->min/max value


dennis00010011b:
### Problem
Min, Max value can be negative
Min can be more or equals than Max 

If you are reporting a problem with Token Wizard, please include the following information:

### Which network did you use? (Mainnet, Kovan, Rinkeby, etc.)

*n/a*

### If you were able to create it, what is the URL of your crowdsale?

*n/a*

### Do you have screenshots showing the problem?

![whitelistminmax](https://user-images.githubusercontent.com/14067096/35953536-fd597fc2-0c39-11e8-9441-47165c8f296a.png)

### Do you see errors in the dev console? If yes, please include a screenshot

To open the dev console in Google Chrome, press F12, or go to `View -> Developer -> Developer Tools`, and open the Console tab

If you see errors, please right click on them and "Save as..". Zip saved file and attach it to the Issue.

*No*

---

PR #659, opened 2018-03-06
(Feature) Validate imported whitelist addresses

### LINKED ISSUE #641: (Improvement) Validate imported whitelist addresses
See #639.

fvictorio:
Depends on #656. Closes #641.

### DIFF
--- a/src/components/Common/WhitelistInputBlock.js
+++ b/src/components/Common/WhitelistInputBlock.js
@@ -6,9 +6,10 @@ import Papa from 'papaparse'
 import '../../assets/stylesheets/application.css';
 import { InputField } from './InputField'
 import { TEXT_FIELDS, VALIDATION_TYPES } from '../../utils/constants'
-import { validateAddress } from '../../utils/utils'
 import { WhitelistItem } from './WhitelistItem'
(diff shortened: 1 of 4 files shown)


vbaranov:
@fvictorio could you clarify which types of validations we cover with this PR? do we need more validations, or it covers all possible validations?

fvictorio:
It performs the same validation that are done when you type an address in the input (it just uses `Web3.util.isAddress`).

vbaranov:
Ok, I see. The issue is that when we import an address with individual min cap > individual max cap there is no validation for now. will it be resolved in the frame of https://github.com/poanetwork/token-wizard/issues/578 ?

fvictorio:
Oh, no, the min an max cap are not being validated (it just checks that they are numbers). We should think of a way to have those validations in a single place and use that both when importing a CSV and when typing an address.

Do you think we should do that as part of this PR?

vbaranov:
>Do you think we should do that as part of this PR?

No. I think we could merge it. I mentioned here just to not forget to do this as a part of #578 or as a separate issue. @fvictorio please create a new issue if you prefer to do it on a separate one.

Issue #536, opened 2018-01-30
(Refactor) Consider removing all checks for contract type


fvictorio:
The codebase has several places where it checks the "contract type" selected in step one (I guess crowdsale type would be a better name). But right now only one crowdsale type is supported. Plus, if another type is added in the future, I seriously doubt that having conditionals sprinkled everywhere is going to be the best approach.

So I think all these conditionals only make the code harder to maintain without giving any future benefit. I propose removing them, since 1) it's a fairly low risk change and 2) it's going to significantly increase the readability of the code.

@vbaranov @igorbarinov Thoughts on this?

vbaranov:
Yes, let's remove these obsolete conditions

PR #672, opened 2018-03-08
(Fix) Add address to whitelist in manage screen

### LINKED ISSUE #596: (Bug) Modifiable crowdsale: incorrect behavior if user tries to add existing whitelisted address

fernandomg:
Closes #596

In this PR `whitelistElements` collection was removed and only uses `whitelist`.

Style was applied for the whitelist elements in the manage screen, so the user can recognize what addresses are being modified, added or untouched.

image

Things to fix that I found: 
1. The way how duplicated addresses are being checked. As the same address can have upper and lower cases, this may lead errors.
2. `min` and `max` values are not being checked, so a user can enter a `min` value greater than `max`. The tx will fail, but it won't be clear why.


src/stores/TierStore.js
```diff
@@ -168,33 +168,60 @@ class TierStore {
...
+    whitelist.push(newItem)
+    this.sortWhitelist(crowdsaleNum)
+  }
+
+  @action sortWhitelist = (crowdsaleNum) => {
+    this.tiers[crowdsaleNum].whitelist = this.tiers[crowdsaleNum].whitelist.sort((prev, curr) => {
+      const currentAddress = curr.addr.toLowerCase()
+      const previousAddress = prev.addr.toLowerCase()
+
+      return currentAddress > previousAddress ? -1 : currentAddress === previousAddress ? curr.stored ? 1 : -1 : 1
```
fvictorio: This needs tests.

src/stores/TierStore.js
```diff
@@ -168,33 +168,60 @@ class TierStore {
...
+    })
   }
 
   @action removeWhitelistItem = (whitelistNum, crowdsaleNum) => {
-    let whitelist = this.tiers[crowdsaleNum].whitelist.slice()
-    whitelist[whitelistNum].deleted = true
-    this.setTierProperty(whitelist, 'whitelist', crowdsaleNum)
+    const removedItem = this.tiers[crowdsaleNum].whitelist.splice(whitelistNum, 1)[0]
```
fvictorio: This needs tests too.

src/components/Common/WhitelistInputBlock.js
```diff
@@ -152,15 +152,12 @@ export class WhitelistInputBlock extends React.Component {
             />
           </div>
         </div>
-        {whitelistElements && whitelistElements.map(e =>
+        {whitelist.length > 0 && whitelist.map((item, index) =>
```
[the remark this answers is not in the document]
fernandomg: fixed

src/stores/TierStore.js
fernandomg: done

src/stores/TierStore.js
fernandomg: done, and added test for the other modified/created methods

fernandomg:
@vbaranov conflict resolution done

fvictorio:
@vbaranov Can we merge this?

vbaranov:
@fvictorio formally yes, but if you could give me some hours, I'll appreciate it. I plan to test it too soon.

fvictorio:
Sure, no problem. I just want to merge as much PRs as possible so that I can start working on #536 with less risk.

PR #680, opened 2018-03-12
(Refactor) Add react-web3

### LINKED ISSUE #647: (Feature) Add warning and pause deployment if user accidentally change MetaMask account while deployment
Inspired by https://poanetwork.slack.com/archives/G8HRTP6JZ/p1519910256000437
###  Problem:  
if user accidentally change the metamask account while deployment the deployment continues, option "skip transaction" is  available for next 3-4 transaction and deployment freezes even if user returns back to right account.
###  Possible solution: 
  1. Should be warning : "Not owner. Unable to continue ..........etc."
  2. Deployment should be paused until user switches to right Metamask account.

*Rinkeby*

*https://wizard.poa.network/invest?addr=0x719c2571Af05B64a8c8f17E69c7720BC727368AD&networkID=4*

*https://drive.google.com/open?id=1NNIZsGv3hZMzmXRIk9tdg_90zBErkpjs*

*No*

---

fvictorio:
This PR adds a custom implementation of the functionality and interface provided by [react-web3](https://github.com/coopermaruyama/react-web3).

This refactor is necessary to fix #647, since it adds the `onChangeAccount` callback.

I didn't use `react-web3` because it does not support web3 1.0 yet. Apparently, there is a published version in NPM that does support it, but it didn't work for me. See https://github.com/coopermaruyama/react-web3/issues/25

My idea is to replace this custom implementation with the one in that repo when the issue is fixed, since that component has more features and it's been used and tested.

Besides that:

- I  replaced some usages of `web3store` with the `web3` instance that the component passes by context. I think this is cleaner, and maybe we should aim to remove `web3store` entirely, since it makes a lot of things less testable.

- This doesn't render *any* page of the wizard if `web3` is not present. This is different to what we had before, where several components would check if web3 was present, and would show a warning if not. The invest page is the only problem, since (if I remember correctly) it was supposed to be usable without web3. If that's the case, this PR will break that.


vbaranov:
@fvictorio it should be reverted and finalized: invest page should work without Metamask. Please create PR to revert changes.

fvictorio:
@vbaranov Reverting this is going to be hard, but I can make a quick fix for allowing to use the invest page without Metamask.

PR #685, opened 2018-03-13
(Feature) Warning on account switch

### LINKED ISSUE #647: (Feature) Add warning and pause deployment if user accidentally change MetaMask account while deployment

fvictorio:
Depends on #680. Closes #647.

[Demo](https://imgur.com/a/ezOnn)

This only happens in the deployment modal. You can switch accounts if the deploy didn't start yet.


fvictorio:
@dennis00010011b Can you check if the failing e2e scenarios are correct, or if the problem is with the tests?

fvictorio:
You are right, it doesn't seem to be working now. Maybe some merge break it. I will take a look.

fvictorio:
@vbaranov It should be working now.

I also fixed two unrelated issues:

- The reload wasn't working anymore: if you refreshed the page, it would re-start the deploy from the first transaction.
- If you switched accounts, and then refreshed, the deploy would resume with the invalid account. The only issue with this is that in this case you have to switch to the correct account and then restart again for the deploy to continue.

PR #693, opened 2018-03-14
(Fix) Prevent deployment if minCap isn't valid

fernandomg:
Closes #669

PR #695, opened 2018-03-15
Revert "(Fix) Prevent deployment if minCap isn't valid"

vbaranov:
Reverts poanetwork/token-wizard#693


vbaranov:
@fernandomg it prevents to go to deployment, if global min cap is empty

PR #698, opened 2018-03-15
(Fix) e2e tests with current changes on local deployment in Travis CI

vbaranov:
Relates to https://github.com/poanetwork/token-wizard-test-automation/pull/17
e2e tests with current changes on local deployment in Travis CI


.travis.yml
```diff
@@ -6,7 +6,6 @@ install:
   - npm install
 
 before_script:
-  - npm run installWeb3
```
fvictorio: This may break the unit tests.

fvictorio: Well, this will work because `npm test` runs the e2e tests first, and the `e2e-prepare-start` script installs web3. This makes the tests order-dependent, but it's not a big deal.

vbaranov: @fvictorio you are right. It is better to leave in .travis.yml. I'll redefine `e2e-prepare-start` script instead.

vbaranov:
@dennis00010011b please let us know when e2e tests will be fixed. Then we will be able to merge this PR. And after it, e2e tests will be launching on the version of Wizard in PR, not on production link.

PR #703, opened 2018-03-15
(Chore) Add HMR

fvictorio:
Adds support for Hot Module Replacement. This allows you to change code and see the result without reloading the app (this is specially useful when you are working in step 2 or 3, because otherwise you have to start from the beginning).


fvictorio:
Btw, this only preserves the stat in the stores. State local to a component is not kept between changes.

fvictorio:
@vbaranov Are you OK with this? It does change the development flow a little, sometimes for worse (you may need to manually refresh in some cases, for example if you modified a store). But I think the pros overweight the cons.

PR #705, opened 2018-03-15
(Feature) Make back button work

### LINKED ISSUE #637: (Feature) add *back* button to switch to previous step before deployment started
**Problem**:  A user has no supported ability to return to the previous step before deployment. Some use cases: 
- user decided to reserve more tokens at crowdsale setup step
- user decided to change token ticker, name at crowdsale setup step

**Solution**:  Add *back* button on the left of *continue* button for second and third steps of wizard

fvictorio:
Depends on #703. Part of #637.

Allow user to move between step two and three. I tested going back to step one too and it seemed to work, but I didn't test that so extensively. Going back from step four probably doesn't make sense (and it should show the "deploy on progress" warning anyway).

I caught at least two problems, but I think they can be fixed as separate issues:

- In step 2, if you add reserved tokens, the decimals input is disabled, as expected. But if you continue to step 3 and then go back, the decimals input is enabled.
- If you enter a valid rate in some tier in step 3, going back to step 2 and forward again shows the error message below the rate input.

Neither of these steps makes the deploy fail: you can continue anyway and it will work fine.


fvictorio:
You are right, we should add it. It's very easy, so I'll do it in this PR.

fvictorio:
Ok, change of plans: I'd like to change several things around the `Continue` buttons (for example, how they are disabled, and how the arrow is added, among other things), so I'll make those changes and add the `Back` button next to the `Continue` button in step two and three.

Please merge this if you are OK with that.

fvictorio:
(I meant: I will add those things in another PR)

PR #722, opened 2018-03-22
(Fix) e2e tests with ganache-cli

vbaranov:
Relates to https://github.com/poanetwork/token-wizard-test-automation/pull/19#pullrequestreview-106223631

@dennis00010011b implemented e2e tests for local deployment of token wizard with ganache-cli

- increase in e2e tests speed 3 times
- `npm run e2e-deployRegistry` - deploys Registry contract to local chain and generates `.env` file
- since we already start ganache-cli for smart-contracts unit tests, we still use it for e2e tests too (I replaced order: smart-contract tests first and then e2e tests for now)


fvictorio:
This looks fine, but it reminds me of something I want to add: https://github.com/kentcdodds/nps

This would allows us to organize our npm scripts and also it would make it easier to parallelize stuff (for example, each contract generation could be done in parallel) for faster build runs.

If you are OK with that, I will create an issue and work on it.

vbaranov:
@fvictorio good idea. I totally support.

PR #732, opened 2018-03-26
(Chore) Add nps to replace npm scripts

### LINKED ISSUE #723: (Chore) Use nps to improve npm scripts
Use [nps](https://github.com/kentcdodds/nps) so that npm scripts are easier to read and modify, and see which of them can be parallelized.

See https://github.com/poanetwork/token-wizard/pull/722#issuecomment-375420840

fvictorio:
Closes #723.

This PR moves most npm scripts logic to `package-scripts.js`, and delegates the execution to `nps`.

This has two big advantages: the `package.json` is less bloated, and it's easier to run stuff concurrently. For example, `npm run build` takes 1m30s right now, but after this change the time is reduced to 30s (three times faster) because it parallelizes both the generation and the compilation of the scripts.

I also deleted all the `Win` suffixed scripts: when I tested them, they weren't working on windows anyway. And with this change, if we eventually need to support windows, it's going to be a lot easier (probably a combination of using `path` and `cross-env`). The `config` object in the `package.json` was also removed; that constants are in `package-scripts.js` now.


fvictorio:
@vbaranov I'm not sure. I remember trying to generate the contracts (see this comment https://github.com/poanetwork/token-wizard/pull/382#issuecomment-349938657) in a windows machine, and that they weren't working. But if those instructions were tested by someone on windows, then I must've done something wrong.

I'm going to undo that deletion.

fvictorio:
As far as I can tell, the only difference with the `Win` suffixed scripts is that those use `%npm_package_config_foo%` instead of `$npm_package_config_foo`.

But since I don't use the env config variables anymore in `package-scripts.js`, they should work the same way.

I also noted that the `startWin` script wasn't generating and compiling the `Registry` contract. Since we didn't receive any issues around that, my guess is that nobody has tried to use the wizard on windows, at leaast since the Registry was added.

Anyway, I think just updating the wiki and change `npm run startWin` with `npm start` should be enough.

vbaranov:
@fvictorio ok, I have created a related issue for updating documentation https://github.com/poanetwork/token-wizard/issues/751.

Issue #711, opened 2018-03-16
(Chore) Properly handle numbers using BigNumber library


fernandomg:
I'm creating this issue as a consequence of several bugs being raised due to the numbers handling across the wizard.

The idea is to:
1. eliminate duplicated operations,
1. cast all numeric values to be stored in contracts to `String` type,
1. cast all numeric values to be retrieved from contract to `BigNumber`/`String`,
1. move every calculated value to stores,

Steps to accomplish this task:
- [x] Create a utility with all the necessary calculation
- [x] Update stores to host values
- [x] Refactor `utils/utils.js`
- [x] Refactor `manage/utils.js`
- [x] Refactor `stepTwo/index.js`
- [x] Refactor `stepThree/index.js`
- [x] Refactor `stepFour/index.js`
- [x] Refactor `stepFour/utils.js`
- [x] Refactor `crowdsale/index.js`
- [x] Refactor `crowdsale/utils.js`
- [x] Refactor `invest/index.js`
- [x] Refactor `stores/utils.js`

A list of bugs related to this chore:
1. https://github.com/poanetwork/token-wizard/issues/708
1. https://github.com/poanetwork/token-wizard/issues/706
1. https://github.com/poanetwork/token-wizard/issues/694
1. https://github.com/poanetwork/token-wizard/issues/674
1. https://github.com/poanetwork/token-wizard/issues/671
1. https://github.com/poanetwork/token-wizard/issues/632

---
Please, feel free to update this issue to add any necessary detail or checkpoint required to properly accomplish this task.

fernandomg:
This issue served as a reference for the `final-form` migration of the whole wizard. Despite `ReservedTokens` and `WhitelistAddresses` everything was migrated.

PR #767, opened 2018-04-04
(Refactor) Use final form step three

fernandomg:
This is another step towards the completion of #711 

In this PR `StepTree` was moved entirely into `final-form`, with the exception of whitelist which is now part of `stepThreeForm` (similar to what was done with `reservedTokens` in `stepTwo`).

I didn't manage to prevent error messages for `rate` and `supply` for the Tiers added after the first one. This won't affect the UX, but will display the errors by default.

Besides I've created an `Error` component, which makes easier to display errors related to `final-form` Fields.

This looks still as a WIP, but the idea is to keep moving toward the adoption of `final-form` as fast as possible.


fernandomg:
@dennis00010011b  this PR sligtly changed the DOM structure of the Step Three, and is giving errors with e2e tests.

dennis00010011b:
@fernandomg  I've updated e2e script, test for upload CSV should work now.
Try to update submodule token-wizard-test-automation and commit again

src/components/stepThree/GasPriceInput.js
```diff
@@ -0,0 +1,90 @@
...
+        {this.state.isCustom ? (
+          <input
+            type="number"
+            className="input"
+            value={this.state.customGasPrice}
+            onChange={(e) => this.handleCustomGasPriceChange(e.target.value)}
+          />
+        ) : null}
+        <p className="description">Slow is cheap, fast is expensive</p>
+        <Error name="gasPrice"/>
```
fvictorio: The name shouldn't be hardcoded here, right?

src/components/Common/Error.js
```diff
@@ -0,0 +1,32 @@
...
+const defaultErrorStyles = {
+  color: 'red',
+  fontWeight: 'bold',
+  fontSize: '12px',
+  width: '100%',
+  height: '20px',
+}
+
+export const Error = ({ name, errorStyle }) => (
+  <Field
```
fvictorio: I'm not sure I understand this. Is the `Field` component really necessary? Couldn't we just receive the whole `meta` object from props and use that?

If I'm understanding this correctly, this means that we'll have two `Field`s with the same `name` attribute, but one of them is used only to show errors. It doesn't sound right to me, is there a good reason for this approach?

src/components/Common/TierBlock.js
```diff
@@ -0,0 +1,197 @@
...
+                label={SUPPLY}
+                description={DESCRIPTION.SUPPLY}
+              />
+              {
+                /*
+                  * TODO: REVIEW. I'm not sure about this approach.
+                  * But it worked for me to keep the error messages properly updated for the minCap field.
+                  */
+              }
+              <Field name="minCap" subscription={{}}>
```
fvictorio: Same here thing (as in the Error component). Is the Field necessary? Can't we use the [`change`](https://github.com/final-form/react-final-form#change-name-string-value-any--void) function?

src/components/stepThree/GasPriceInput.js
fernandomg: fixed

fernandomg:
@dennis00010011b I've been running the e2e locally and it is still failing.

This are the lists of issues that I found:
1. For the error messages you use the xpath selector and look for the only attr you had available that is `style`, I'd slightly changed the styles for the `stepThree` and now it's failing. As a result I've just added the `class="error"` to the errors `<p>` tags.
1.  `Wizard step#3: User is able to bulk delete all whitelisted addresses` is failing for me, despite I reproduced the steps in the browser and worked properly.
1.  All test that contains the assert `assert.equal(flagCrowdsale, true);` fail in the `testSuite1.js` (you can see them here: https://travis-ci.org/poanetwork/token-wizard/builds/362808731#L2724)

---

As a note, with this migration to `final-form` we're adding `name` attributes to most of the inputs in the wizard, so it will be easier and cleaner for you to look for a particular tag despite its location in the DOM.

If you find yourself using XPath to reach a particular element, don't hesitate in request a change in the code to add an `id`, a `name` or a `class`, or whatever attribute is needed, to make it easier for you to build the tests.

dennis00010011b:
@fernandomg  Thank you for feedback.
1. I’ll change finding the errors by class name. It would really helpful if elements would have id or at least name or class.
2. I see this issue with bulk delete, still can’t figure out it 
3. Almost all test use pre-created crowdsale . flagCrowdsale used for skip  tests if crowdsale hasn’t created.

PR #783, opened 2018-04-10
(Fix) Remove unnecessary validation

### LINKED ISSUE #781: (Bug) User can't continue if decimals has default value (zero)
#### Step for reproduce:

1. Open https://wizard.oracles.org/,  go to Step2
2. Fill out fields Name, Ticker with valid data
3. Click button Continue

#### Expected result:
 - since field Decimals has value 0 user should be able to proceed to Step 3

#### Actual result:
 - button Continue non-clickable, error message "This field is required" presents

Sokol

n/a

screen shot 2018-04-10 at 08 55 24

*No*
---

fernandomg:
Closes #781


fvictorio:
Uhm, this works, but if the input is empty the error message is `Should not be greater than 18`.

PR #833, opened 2018-04-26
(Feature) Reserved tokens in manage page

### LINKED ISSUE #811: (Feature) display reserved tokens destinations in management page of Wizard
**Problem**: When users create crowdsale and forget to click "+" button, they think that they added reserved tokens destinations, but actually the didn't.

**Solution**: To simplify the process of the check, whether tokens are reserved or not, let's display reserved tokens destinations block on the management page of Token Wizard. The block should be read-only and visible only for the crowdsale owner.

fernandomg:
Closes #811

List of reserved tokens is listed inside the Distribute reserved tokens block.

Maybe, a better idea would be to add a button to trigger a modal with an scrollable list of addresses.

If you prefer that, I can implement it for the `v2.0`,  and then merge the changes back to `v1.0`


fernandomg:
@vbaranov I didn't see your message until now.

I've tried to reproduce, but I'm getting all the reserved addresses (tokens and percentage)/

image

vbaranov:
@fernandomg I forgot to say, where I deployed this crowdsale. It is in Sokol. And the problem still exists. Please, look at my screenshot of manage page:

screen shot 2018-05-03 at 16 38 56

But actually tokens are reserved in perecents too for the same user:

screen shot 2018-05-03 at 16 40 59

So, I reserved tokens and percents for the same user. Maybe, this is the issue?

fernandomg:
> So, I reserved tokens and percents for the same user. Maybe, this is the issue?

@vbaranov indeed.

If you see the code in the contract, it will assign the latest value entered to the address.

https://github.com/poanetwork/ico/blob/cd5301aa90f38f60b0d038d895e5b9435d622272/contracts/MintableTokenExt.sol#L129

vbaranov:
@fernandomg could we display tokens for the same user in 2 lines? one line for reserved tokens and one line for percentage?

fernandomg:
@vbaranov in fact, if you add `0x123`, `percentage`, `65`; and after it `0x123`, `tokens`, `150`. Percentage will be overwritten by tokens, and the last will remain. Both are not supported for the same address.

fernandomg:
@vbaranov oh, oh... sorry for the missunderstanding. I'll fix rhat ASAP

fernandomg:
@vbaranov finally!

image

PR #880, opened 2018-05-15
Proper token bough based on hard cap

### LINKED ISSUE #872: (Bug) wrong amount of tokens in success alert if a user tries to buy more than hard cap of a tier
**Problem**: the same as #847, but when a user tries to buy more than a hard cap of a tier

**Solution**: check remaining tokens to buy in a tier and amount, that user is trying to buy and show the lowest value from those in the message.

fernandomg:
Closes #872

I also implemented the hard cap limit for the Dutch Auction strategy


fernandomg:
@vbaranov fixed!

PR #916, opened 2018-05-25
(Fix) Display warning at invest page if user is not whitelisted

vbaranov:
Display warning at invest page if user is not whitelisted or he is whitelisted but he have bought all permitted amount of tokens

screencapture-localhost-3000-invest-2018-05-25-18_02_40


dennis00010011b:
This issue in branch 2.0 . https://github.com/poanetwork/token-wizard/issues/917

PR is ok

PR #935, opened 2018-06-01
(Fix) Rename invest

vbaranov:
Rename "invest" to "contribute" where it is not renamed: 
- contribution page path
- React Component
- MobX storage for contribution
- some variables and constants
- some titles and descriptions


vbaranov:
@dennis00010011b could you take a look, whether e2e tests are affected? e2e tests for both strategies are failed. I've changed the names of some css classes for invest page as well.

PR #954, opened 2018-06-12
(Fix) Allow min cap edit after crowdsale started

### LINKED ISSUE #948: (Fix) changing of global min cap for ongoing crowdsale
**Problem**: global min cap is alternative to whitlists functionality. Whitelists can be changed any time before tier ends. Global min cap should work in the same manner.

**Solution**: make global min cap modifiable before and for ongoing crowdsale if all of the tiers are not whitlisted.

fernandomg:
Closes #948


dennis00010011b:
@fernandomg 
minCap field is modifiable after crowdsale time expired. Should be disabled
screen shot 2018-06-12 at 08 25 46

fernandomg:
> will it be enabled, if crowdsale is not modifiable? It should because changing of whitelists and global min cap doesn't depend on isModifiable flag

ok, just added that condition

fernandomg:
Ok, I refactored a bit the manage page startup flow. Now flags used to validate editing are centralized and more explicit. Hope this helps avoid undesired outcomes.

@dennis00010011b  @vbaranov

dennis00010011b:
@fernandomg 
disabled to contribute new  minCap value. Field 'Contribute'  is validated with old minCap
screen shot 2018-06-13 at 15 27 14

fernandomg:
@dennis00010011b I don't think this is a bug related to the manage screen, as you're able to see the value has properly changed.

Despite that, I tried to reproduce it but with no luck. MinCap is updated properly in the invest page.

Please check: `0xf0d642dbc7517672e217238a2f008f4f8cdad0586d8ce5113e9e09dcc6860619`

Issue #957, opened 2018-06-13
(Fix) configure Token Wizard 2.0 to support Auth-os Proxy for crowdsales


vbaranov:
Configure Token Wizard 2.0 to support Auth-os Proxy contracts for crowdsales:
https://ropsten.etherscan.io/address/0x86501d6bbe1e876db117e313ed7bcf58691cf378#code

PR #958, opened 2018-06-14
(Refactoring)  Support of Auth-os Proxy contract: part 1

vbaranov:
Relates to https://github.com/poanetwork/token-wizard/issues/957

First part of integration with Auth-os Proxy contract - deployment of Proxy and partially getters for `crowdsale` and `contribute` pages
Nothing is changed in UI.

### DIFF
--- a/src/components/contribute/index.js
+++ b/src/components/contribute/index.js
@@ -21,7 +21,7 @@ import {
   getUserMaxLimits,
   getUserMinLimits
 } from '../crowdsale/utils'
-import { countDecimalPlaces, getExecID, getNetworkID, toast } from '../../utils/utils'
+import { countDecimalPlaces, getExecID, getAddr, getNetworkID, toast } from '../../utils/utils'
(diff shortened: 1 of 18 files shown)


dennis00010011b:
@vbaranov 
Reserved addresses are not supported
screen shot 2018-06-14 at 09 03 40

dennis00010011b:
Also multitiers initialization skipped

fernandomg:
> Reserved addresses are not supported

Same. It worked for me after removing `tokenManagerMintedCapped` param from: https://github.com/poanetwork/token-wizard/blob/122224d92e580c52dd4d8823b9b6be9d0006c5a2/src/components/stepFour/utils.js#L398

fernandomg:
> Also multitiers initialization skipped

Same as with reserved tokens but with `saleManagerMintedCapped` in:
https://github.com/poanetwork/token-wizard/blob/122224d92e580c52dd4d8823b9b6be9d0006c5a2/src/components/stepFour/utils.js#L491

PR #963, opened 2018-06-18
(Fix) Deployment modal columns

### LINKED ISSUE #947: (Fix) show multiple tiers columns only for adding whitelists for tiers in deployment modal
**Problem**: 

screen shot 2018-06-11 at 15 50 14

Only addition of whitelists is executing per tier. We don't need to show multiple columns if crowdsale is not whitelisted.

**Solution**: show multiple tiers columns in deployment modal if any tier other than first is whitelisted

fernandomg:
Closes #947

Tier column will be listed only if that particular tier is whitelisted. First tier is always shown despite it's whitelisted or not.

image

---
@dennis00010011b @vbaranov One issue I'm facing is that I cannot deploy the crowdsale with a setup like the one in the screenshot: 4 Tiers, first and 3rd not whitelisted, or even 4 tiers all whitelisted. It's always failing in the `Add tiers to Crowdsale` step. Have you experienced a similar issue?


vbaranov:
@fernandomg great catch! Yes, I repeated the same issue even without whitelists.

PR #964, opened 2018-06-18
(Fix) Get length of all applications

vbaranov:
Get the length of all applications from Auth-os to display statistics. The statistics page is available at `/stats`


dennis00010011b:
@vbaranov  
1.Amount of dutch crowdsales didn't changed after such crowdale was created and finalized
2. There is an error message if network is not Ropsten
screen shot 2018-06-18 at 20 08 01

vbaranov:
@dennis00010011b your crowdsale is accounted in stats:
screen shot 2018-06-19 at 09 34 14

And I can't reproduce the error on networks switching. I'm gonna to merge this PR. If problem with networks switching will repeat please open an isuue for that.

PR #970, opened 2018-06-19
(Fix) Dutch auction crowdsale: is_whitelisted flag

vbaranov:
Auth-os team has implemented `is_whitelisted` output for `getCrowdsaleStatus` function. It is implemented in TW 2.0 in current PR.

Also,
- `getTokensSold` function is also implemented in Auth-os for Dutch Auction (try/catch is removed)


src/components/manage/index.js
```diff
@@ -372,22 +372,16 @@ export class Manage extends Component {
 
       } else if (isDutchAuction) {
         const tier_data = await getCrowdsaleStatus(...params).call()
+        if (tier_data && !tier_data.hasOwnProperty('whitelist_enabled')) {
+          tier_data.is_whitelisted = tier_data[5]
+        }
         const tier_dates = await getCrowdsaleStartAndEndTimes(...params).call()
         const crowdsaleWhitelist = await getCrowdsaleWhitelist(...params).call()
         const num_whitelisted = crowdsaleWhitelist.num_whitelisted || crowdsaleWhitelist[0]
```
fernandomg: no longer used, can be removed

src/components/manage/index.js
```diff
@@ -372,22 +372,16 @@ export class Manage extends Component {
 
       } else if (isDutchAuction) {
         const tier_data = await getCrowdsaleStatus(...params).call()
+        if (tier_data && !tier_data.hasOwnProperty('whitelist_enabled')) {
+          tier_data.is_whitelisted = tier_data[5]
```
fernandomg: `tier_data.whitelist_enabled` should exist, as the `crowdsaleData` method is expecting it instead of `is_whitelisted`.

https://github.com/poanetwork/token-wizard/blob/874209ae06233b4c1de4b54c44717ca557795013/src/components/manage/utils.js#L162

vbaranov: @fernandomg, right, fixed
@dennis00010011b the issue with whitelist is unrelated to this PR, but to the changes in Auth-os applications for TW. I guess It should be fixed in a separate PR.

Issue #978, opened 2018-06-25
(Bug) Minted-Contribution page: incorrect text in mincap field if crowdsale has ended


dennis00010011b:
Expected result: text "You are not allowed"
Actual result: text "Infinity test"

If you are reporting a problem with Token Wizard, please include the following information:

### Which network did you use? (Mainnet, Kovan, Rinkeby, etc.)

*Ropsten*

### If you were able to create it, what is the URL of your crowdsale?

* 0x19628ef4f1e52435a91694e0b17e2591e7982b96f1883cadfe58900fe74230c4*

### Do you have screenshots showing the problem?

*<img width="1440" alt="screen shot 2018-06-25 at 13 15 06" src="https://user-images.githubusercontent.com/14067096/41873444-d91b39b4-7879-11e8-8ca6-013ed2322e72.png">
*

### Do you see errors in the dev console? If yes, please include a screenshot

No

---

Issue #977, opened 2018-06-25
(Bug) Minted -Contribution page: Incorrect minCap value is displayed (for not whitelisted tiers)


dennis00010011b:
Steps for reproduce:
1. Create Minted crowdsale with mincap not equals zero, whitelist disabled.
2. Observe minCap on contribution page.

Expected result:
 - displayed mincap should be equal  the given value

Actual result:
 - displayed mincap is 0.000000000000001
 - mincap validation works properly

If you are reporting a problem with Token Wizard, please include the following information:

### Which network did you use? (Mainnet, Kovan, Rinkeby, etc.)

*Ropsten*

### If you were able to create it, what is the EXEC ID of your crowdsale?

*Exec ID : 0x42b680591beb20d1de0d4ca368d3f10acc6db07bc74b68a9681655fb351b9039*

### Do you have screenshots showing the problem?

<img width="1440" alt="screen shot 2018-06-25 at 12 25 38" src="https://user-images.githubusercontent.com/14067096/41871620-4db966f2-7874-11e8-84eb-5f0c63d5c44c.png">

### Do you see errors in the dev console? If yes, please include a screenshot

No
---

PR #983, opened 2018-06-26
(Fix) display correct min cap in contribution page

vbaranov:
Closes https://github.com/poanetwork/token-wizard/issues/977
Closes https://github.com/poanetwork/token-wizard/issues/978

display correct min cap in contribution page

### DIFF
--- a/src/components/crowdsale/utils.js
+++ b/src/components/crowdsale/utils.js
@@ -358,13 +358,16 @@ const getRate = async (addr, execID, methods) => {
 
 const calculateMinContribution = async (method, decimals, naturalMinCap, isWhitelisted) => {
   //todo: update for Proxy
-  const { minimum_contribution, minimum_purchase_amt, max_tokens_remaining, max_purchase_remaining } = await method.call()
-  const minimumContribution = toBigNumber(minimum_contribution).times(`1e-${decimals}`)
+  const { tier_min, minimum_contribution, minimum_purchase_amt, max_tokens_remaining, max_purchase_remaining } = await method.call()
+  const minimumContributionDutch = toBigNumber(minimum_contribution).times(`1e-${decimals}`)
+  const minimumContributionMintedCapped = toBigNumber(tier_min).times(`1e-${decimals}`)
+  const minimumContribution = minimumContributionMintedCapped || minimumContributionDutch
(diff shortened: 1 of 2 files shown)


dennis00010011b:
@vbaranov 
This issue still reproducible for Dutch
`0xbd5c86c9fbaa47a598cf55ad1cee7a1d8f8450857cb10e0f122503b8c0e65243 `
 Ropsten
screen shot 2018-06-26 at 13 26 14

src/components/crowdsale/utils.js
```diff
@@ -358,13 +358,16 @@ const getRate = async (addr, execID, methods) => {
 
 const calculateMinContribution = async (method, decimals, naturalMinCap, isWhitelisted) => {
   //todo: update for Proxy
-  const { minimum_contribution, minimum_purchase_amt, max_tokens_remaining, max_purchase_remaining } = await method.call()
-  const minimumContribution = toBigNumber(minimum_contribution).times(`1e-${decimals}`)
+  const { tier_min, minimum_contribution, minimum_purchase_amt, max_tokens_remaining, max_purchase_remaining } = await method.call()
+  const minimumContributionDutch = toBigNumber(minimum_contribution).times(`1e-${decimals}`)
+  const minimumContributionMintedCapped = toBigNumber(tier_min).times(`1e-${decimals}`)
+  const minimumContribution = minimumContributionMintedCapped || minimumContributionDutch
```
fernandomg: Still is returning `naturalMinCap` as `toBigNumber` is returning a bigNumber object with value `0` when the parameter passed is not a number or a string representation of a number.

A solution would be to replace lines `362..364` with:
```
const minimumContribution = toBigNumber(tier_min || minimum_contribution).times(`1e-${decimals}`)
```

But I think that after this, in a different PR, we should replace `toBigNumber` signature and return `undefined` when an invalid value is passed.

What you think @vbaranov ?

vbaranov: Yes, agree. It would be nice to mark invalid values as `undefined`

vbaranov:
@fernandomg @dennis00010011b fixed, thank for your suggestions. Also, I've added a fix for https://github.com/poanetwork/token-wizard/issues/978, as it is related to the same page component.

PR #991, opened 2018-06-28
(Feature) toBigNumber signature change

### LINKED ISSUE #985: (Feature) Modify toBigNumber signature
If the parameter passed to the method is not a representation of a number, then avoid returning a BigNumber instance, return `undefined` instead.

fernandomg:
Closes #985


fernandomg:
@vbaranov done. Moved it to `utils/utils` and also used `toBigNumber` consistently across the whole project.

PR #1011, opened 2018-07-04
(Feature) Add Burn excess support

### LINKED ISSUE #997: (Feature) Burn flag for Dutch Auction
**Problem**: Crowdsale owners have no ability to manage what to do with unsold amount in Dutch Auction. Unsold amount is locked after crowdsale finalization.

**Solution**: Auth-os team has implemented a flag `_burn_excess` for Dutch Auction. Let's implement this flag in UI.
Tasks:
- add horizontal rario buttons group "Burn unsold" in step 3 in "Global settings" section with values "Yes" and "No". 
- value of the flag should be mapped to `burnExcess` variable: https://github.com/poanetwork/token-wizard/blob/53a682b8e77f1c9a494a50a1ee81fac2ea19971c/src/components/stepFour/utils.js#L184
- the value of burn flag should be added to the summary file "DutchCrowdsale_data.txt" in `*******CROWDSALE SETUP*******` section. For example: `Burn unsold amount: yes`
- the value of the flag should be added to Publish page as well
- the value of burn flag should be added to management page to the disabled input (there is no way to change the value of this flag) at the last position in the page

fernandomg:
Closes #997

Everything was updated as stated in the issue, except that in the `manage` page I added `Burn Excess` to the global info.

image

I think it makes sense to have it along with the `wallet address` as in the StepThree.


dennis00010011b:
@fernandomg 
Owner can't finalize if all tokens sold and time expired
`0x4b4490a7fcc2363cdd930a9da20596c1a1c6f85f0ab1bb09623387f147858975`
`Ropsten`

fernandomg:
@dennis00010011b fixed the `Finalize` button status.

This issue happens in `2.0` branch as well, and has nothing to do with changes made for this PR. Can you confirm?

In fact, as far as I understand, `Finalize` button status has nothing to do with `burnExcess` flag. This will just enforce a _burn_ or _send to the crowdsale walllet_ of the remaining tokens when crowdsale is finalized.

dennis00010011b:
@fernandomg You are right, these issues in branch 2.0

Issue #1012, opened 2018-07-04
(Fix) Replace getCurrentTierInfo with getCrowdsaleTier for minted capped strategy


vbaranov:
**Problem**: `getCurrentTierInfo(...)` function for minted capped crowdsale strategy strongly depends on `block.timestamp`, so, it changes not continuously. This leads returning of incorrect data for a tier until new block will be issued.

**Solution**: Instead, the suggestion is to add a function, that will calculate the current tier index (by iterating through [getTierStartAndEndDates](url) data for every tier) and then will call [getCrowdsaleTier](https://github.com/auth-os/applications/blob/2788cb547122b4fdfc3952037f7cad807c92abe5/TokenWizard/crowdsale/MintedCappedCrowdsale/contracts/MintedCappedIdx.sol#L451) function with this index

Note:
The number of tiers can be retieved with [getTiersLength](https://github.com/poanetwork/token-wizard/blob/2.0/src/components/crowdsale/utils.js#L270) function

PR #1014, opened 2018-07-04
(Feature) Display start and end rates for DutchAuction

### LINKED ISSUE #938: (Feature) display 3 prices for Crowdsale page for Dutch Auction strategy
**Problem**: crowdsale page displays only current price of the token. It could be useful to see start and end prices also.
**Solution**: display start and end prices of crowdsale in the crowdsale page as well

fernandomg:
Closes #938 

For Dutch Auction strategies adds a third row for the different Rates (Start, Current and End)

`Price` was replaced with `Rate`.

image


dennis00010011b:
@fernandomg Contribution page: current rate is the latest value of rate before ending
Should be 0 or text ' --- '
screen shot 2018-07-04 at 08 34 17

vbaranov:
@fernandomg @dennis00010011b 

Yes, good idea. Such rule could be: before crowdsale's start and after crowdsale's end we should display `0` rate. For both strategies for all pages where the current rate is displayed.

fernandomg:
@vbaranov @dennis00010011b 

I've just updated `manage` and `crowdsale` screens.

For the **MintedCapped** crowdsale, I did the change in a per-tier basis (if tier is not active, then `rate` is `0`)

image

dennis00010011b:
@fernandomg 
Manage & Crowdsale page: current rate hasn't changing during the tier;
displayed current rate always equals min rate
duration of crowdsale  = 2 min
screen shot 2018-07-05 at 14 24 56
screen shot 2018-07-05 at 14 24 20

dennis00010011b:
@fernandomg 
3 tiers crowdsale-crowdsale page: rate is always from 1st tier
screen shot 2018-07-05 at 14 50 57

dennis00010011b:
@fernandomg @vbaranov 
Also, if displayed rate = 0 before start then  the  rate value in-fact hidden from investors. They can't know price for future MintedCapped  ICO. Is it relevant to hide info from investors?

fernandomg:
> @fernandomg @vbaranov
Also, if displayed rate = 0 before start then the rate value in-fact hidden from investors. They can't know price for future MintedCapped ICO. Is it relevant to hide info from investors?

Thinking in that way I'm not sure how useful is to display `0` if a tier/crowdsale hasn't started yet. In fact, for Dutch Auction, the contract will return `0` if the crowdsale didn't start yet and the `max rate` when finished.

On the other hand, Minted Capped crowdsale will always return the actual rate for every tier.

---

Makes any sense to _obfuscate_ some information that can be retrieved openly from another media, like querying the contract directly?

fernandomg:
> Manage & Crowdsale page: current rate hasn't changing during the tier;
displayed current rate always equals min rate
duration of crowdsale = 2 min

@dennis00010011b 
Actually, that's something that's the value that the contract is returning and has nothing to do with this implementation. I'll leave you here a progression of rates...

image

image

image

image

It was a 5' Dutch Auction after it started `Current Rate` was in `301`, and it progressed quite fast up to `885`, then it stuck there until the last second when the rate moved to `10959`, and a second afterwards it went to `0`.

fernandomg:
> 3 tiers crowdsale-crowdsale page: rate is always from 1st tier

Indeed, that's because it's querying the method `getCurrentTierInfo`, which I have to change in this PR: https://github.com/poanetwork/token-wizard/issues/1012

In fact, `rate` will be `0` if the crowdsale didn't start or the `rate` that the contract provides.

fernandomg:
> Manage page: can't add whitelisted address , transaction hash't created

@dennis00010011b 
fixed and merged `2.0` to this branch.

vbaranov:
@fernandomg 
>As Dennis said I don't know if it makes any sense to obfuscate some information that can be retrieved openly from another media, like querying the contract directly.

I agree, makes sense to display returning data from smart-contracts. Do you want to make some changes to follow this point? or they are already included in this PR?

fernandomg:
> I agree, makes sense to display returning data from smart-contracts. Do you want to make some changes to follow this point? or are they already included in this PR?

I have to undo/change some of the changes done in order to overwrite values returned by the contract

fernandomg:
- `getCurrentCrowdsaleRate` was removed and its usages
- `rate` value now is displayed as it is in the contracts
- `crowdsale` page now has the legend _rate_ instead of _price_ for both strategies

PR #1018, opened 2018-07-05
(Fix) Update to work with both proxies

vbaranov:
Relates to https://github.com/poanetwork/token-wizard/issues/957

The 4th step of implementing proxies:
- proxies assets were added for both strategies (source codes, ABIs, binary codes)
- registry exec and provider address are updated
- overall fixes to support new implementations of Proxies

Nothing will be changed from UI perspective with this PR


dennis00010011b:
@vbaranov 
Minted-Contribution page
screen shot 2018-07-05 at 10 00 23

Issue #1016, opened 2018-07-04
(Bug)Dutch: button `finalize` disabled if crowdsale time over but not all tokens were sold


dennis00010011b:
Steps:
1. Create Dutch crowdsale, duration 5 min, minCap
2. Buy minCap after start if crowdsale
3. Wait until crowdsale will end
4. As an owner open manage page, observe button `finalize`

Expected result:
 -   button `finalize ` should be enabled since crowdsale has ended

Actual result:
 - button `finalize ` disabled
 - owner can't finalize

If you are reporting a problem with Token Wizard, please include the following information:

### Which network did you use? (Mainnet, Kovan, Rinkeby, etc.)

`Ropsten`

### If you were able to create it, what is the URL of your crowdsale?

`0x26bc4e48e81e55b95390769c693a74ca15d6150354d9f3860707b756ab0b774f`

### Do you have screenshots showing the problem?

<img width="1440" alt="screen shot 2018-07-04 at 11 47 23" src="https://user-images.githubusercontent.com/14067096/42291536-27b526d0-7f81-11e8-9817-14b58634664e.png">

### Do you see errors in the dev console? If yes, please include a screenshot

No

---

PR #1021, opened 2018-07-05
(Fix) make default token visibility 18 decimals 

### LINKED ISSUE #1015: (Fix) Make default token visibility 18 decimals
**Problem**: The current default value for `decimals` field of a token is `0`. Though the most ~~popular~~ recommended divisibility of token is 18.

**Solution**: Let's make `18` decimals by default.

mariano-aguero:
Closes #1015 

- Add decimal constant for token

See image
screenshot_20180705_162459

### DIFF
--- a/src/components/stepTwo/index.js
+++ b/src/components/stepTwo/index.js
@@ -2,6 +2,7 @@ import React, { Component } from 'react'
 import '../../assets/stylesheets/application.css'
 import { checkWeb3 } from '../../utils/blockchainHelpers'
 import { StepNavigation } from '../Common/StepNavigation'
+import { Loader } from '../Common/Loader'
 import { clearingReservedTokens } from '../../utils/alerts'
 import { NAVIGATION_STEPS, VALIDATION_TYPES, initialStepTwoValues } from '../../utils/constants'
 import { inject, observer } from 'mobx-react'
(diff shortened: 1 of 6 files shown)


src/components/stepTwo/index.js
```diff
@@ -21,7 +21,7 @@ export class stepTwo extends Component {
       tokenValues: {
         name: this.props.tokenStore.name,
         ticker: this.props.tokenStore.ticker,
-        decimals: this.props.tokenStore.decimals || 0
+        decimals: this.props.tokenStore.decimals || initialStepTwoValues.token.decimals
```
fernandomg: I'd rather prefer if the change is done directly in the Store:

https://github.com/poanetwork/token-wizard/blob/1151b01502267b85b02e304ab9973f1d40a372f4/src/stores/TokenStore.js#L24

and enforce a consistency. Also, can be a good option to get rid of the `initialStepTwoValues` or use those values to setup all the initial values in the required Stores.

What you think?

mariano-aguero: @fernandomg 
It's ok for me. I think it can be done with some refactoring. We should keep the current default values ​​so we do not generate any problems

mariano-aguero:
I make a refactoring on step 2, now the tokenStore is used, what do you think? Any suggestions?

mariano-aguero:
@dennis00010011b I tried to reproduce the problem without success.
Create this video, I do something different to your steps to get to the error?

https://www.useloom.com/share/cc172ff561aa455db4afb84f8c5197d5

dennis00010011b:
@mariano-aguero 
Sorry, there is no issue. It was something wrong with my local deployment.
One more concern about this PR that it contains closed issue https://github.com/poanetwork/token-wizard/issues/1016

mariano-aguero:
@dennis00010011b No problem, it's your job and you do it very well :) . So the PR is good for you? Please do not close this PR until @fernandomg  reviews it

src/components/stepTwo/index.js
```diff
@@ -18,32 +19,41 @@ export class stepTwo extends Component {
...
 
-  componentDidMount() {
-    checkWeb3(this.props.web3Store.web3)
+  async componentDidMount() {
+    const { tokenStore, web3Store } = this.props
+    await checkWeb3(web3Store.web3)
+
+    if (tokenStore.isEmpty) {
+      tokenStore.addTokenSetup()
+      this.tokenValues = tokenStore.getToken
```
fernandomg: One side effect that concerns me regarding this change, is that it's considering the `tokenStore` as _emtpy_ if the `reservedTokensInput` array is empty, but you can create a token without reserved tokens. Also, Minted Capped strategy does not use `supply` value. 

At first sight, this has a no bigger impact, but how does it behave with the `autosave` feature? If you're in `stepTwo` with token data loaded (without reserved token addresses) and refresh the page, then the token data will be erased. Same will happen if you navigate back from `stepThree`.

Despite _autosave_ is not properly working in the `2.0` branch, I think it should be fixed before the release of TW 2.0 (@vbaranov?)

vbaranov: Yes. Switching between steps in both directions (or refreshing) with losing data can be annoying for end-user. This is why I was creating https://github.com/poanetwork/token-wizard/issues/637. It would be nice if we could save state of the page until deployment process finished.

As about current PR I would suggest some changes. As @fernandomg noticed, token object differs for different strategies. Maybe, let's complement `isEmpty` and `getToken` computed props with the returning/operating with different data depending on current strategy? @mariano-aguero waht do you think?

Basically, token object has 4 properties: name, ticker, decimals, supply (*note*: *initital supply is 0 for minted capped strategy, because token is mintable in this strategy*)
Reserved tokens metadata is extension of token object and available only for minted capped startegy.

mariano-aguero: Hi guys, thanks for the feedback, 

@fernandomg
- Removed reservedTokensInput to validate if it is empty
- The store 'ReservedTokenStore' is used in this step to save these tokens

@vbaranov
- Change how isEmpty and getToken is working like you suggested
- Reserved tokens are maintained in another store

Both:
  - In the case that you return from step 3 to step 2, the data from step 2 will be maintained (I move a line). In the case of a reload of the page, I think we should think about using https://github.com/pinqy520/mobx-persist because the stores are initialized empty

mariano-aguero:
Hi @fernandomg , I really appreciate your feedback. In regard to data persistence, I create an issue #1033 .I think we should work it apart since it would take too much work, and it has nothing to do with the original issue. Thank you for your effort

PR #1046, opened 2018-07-14
(Refactor) Of method getOwnerApplicationsInstancesForProxy

fernandomg:
Refactors `getOwnerApplicationsInstancesForProxy`
 - makes use of async/await syntax
 - sorts the result of contracts from newest to oldest

Check the [discussion about this refactor](https://github.com/poanetwork/token-wizard/pull/1024#discussion_r202238084) for reference

### DIFF
--- a/src/utils/blockchainHelpers.js
+++ b/src/utils/blockchainHelpers.js
@@ -348,34 +348,28 @@ async function getOwnerApplicationsInstancesForProxy() {
   const { web3 } = web3Store
   const proxiesRegistryContract = await attachToSpecificCrowdsaleContract('ProxiesRegistry')
   const accounts = await web3.eth.getAccounts()
+  const execIDs = await proxiesRegistryContract.methods.getCrowdsalesForUser(accounts[0]).call()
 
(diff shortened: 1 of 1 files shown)


src/utils/blockchainHelpers.js
```diff
@@ -348,34 +348,28 @@ async function getOwnerApplicationsInstancesForProxy() {
...
-        if (appNameLowerCase.includes(mintedCapped) || appNameLowerCase.includes(dutchAuction)) {
-          crowdsales.push({ appName, execID: proxyAddr })
-        }
-        resolve()
-      } catch (error) {
-        logger.error(error)
-        resolve()
+  const contracts = []
+
+  for (const execID of execIDs) {
```
vbaranov: @fernandomg if I'm not wrong, in this case, smart-contracts' functions' execution will be made in a sequence, instead of parallel calls (from the current implementation). Thus, suggested implementation is slower for the number of proxy addresses greater then 1, isn't it? What are the benefits of suggested implementation?

PR #1066, opened 2018-08-01
(Fix) Handle hardCap in Contribute Page

### LINKED ISSUE #726: (Fix) Stop countdown if hard cap is reached immediately
Relates to https://github.com/poanetwork/token-wizard/pull/720#pullrequestreview-106585780

fernandomg:
Closes #726

`CountdownTimer` will display a **_Tier sold out_** message on a multitier Crowdsale if the tier sold out is not the last one.

As per @dennis00010011b suggestion in the issue ([see here](https://github.com/poanetwork/token-wizard/issues/726#issuecomment-409352100)) when last tier (or a single tier) is sold out, then message **_Crowdsale has ended_** will be displayed.

Also added tests for the components modified.


fernandomg:
@dennis00010011b can you check once more if the status is still **Tier sold out** for your account?

As I was trying to reproduce this, I've got the same results as you, but after 2 minutes I refreshed and everything worked just fine. Even I was able to buy the tokens for the second tier.

dennis00010011b:
@fernandomg I have waited more than 10 min after tier's start and got same result

PR #1072, opened 2018-08-07
(Fix) Investors incorrect max cap

### LINKED ISSUE #1067: (Bug) Minted-Contribution page: Incorrect maxCap for whitelisted investors which added from CSV file
Steps:
1. Create Minted crowdsale, 1 tier, duration by default, decimals=18, rate=1000, supply=100
2. In Step 3 add whitelisted addresses from file :
https://drive.google.com/open?id=1jGWohlqK3PttrPHT_Sma1ousATHSm0gz
Private keys: https://drive.google.com/open?id=1QB8Y6qm11ZBRbPd9kiRF7ccd1_ALBJPN

3 .Finish deployment
4. Set up  any whitelisted account in wallet
5. Open contribution page, observe max contibution field, try to buy

Expected result:
 - max contribution field should contain maxCap value for current  whitelisted investor
 - whitelisted investor should be able to buy

Actual result:
 - whitelisted address can't buy
 - incorrect value of max contribution

*Ropsten*

*proxyID 0xce4c704c3af8f359929ed09d3fe3739913d5e263*

uuu

No

---

fernandomg:
Closes #1067 

This PR fixes all the related issues with whitelisted addresses and tier's supply. Scenarios that I took into account for both strategies (MintedCapped and DutchAuction):
- Whitelists addresses added (from CSV, manually or combined), will be limited to the sum of maxCap that is not greater than tier's supply. i.e.: for addresses with `maxCap` of `[100, 200, 300]` if `supply` is `400`, then only the first two addresses will be added.
- If after adding whitelist addresses, the `supply` value will be validated against the sum of the `maxCap`s. It should not be less than the `maxCap` sum.
- In the `manage` screen `supply` cannot be edited, so whitelist addresses will be checked.

**Note**
The scenario that I didn't fix, due to the complexity of the fix itself (it would require several modifications and would be better to address it in a separate PR), has to be with `manage` screen and an address values modification.

As you can see in the screenshot
image
Despite having `1011` tokens available, and requiring only `28` extra tokens to change `1972` to `2000`, it's giving the error as it considers that the address is new and not a modification of an existing one.

---
Also:
- tests where updated and fixed
- linting errors where solved


dennis00010011b:
@fernandomg 
Manage page: allowed to add whitelisted investors with maxCap greater than remained token's supply. Steps:
1. Create Minted crowdsale, 1 tier:
  - supply =1000
 - whitelisted address with maxCap=300
2. As whitelisted investor buy maxCap
3. As an owner open manage page
4. Add another whitelisted investor with maxCap=supply

Expected result: 
 - whitelisted maxCap should be restricted by supply-300=700

Actual result:
 - it is possible to add whitelisted address wjth maxCap >700
screen shot 2018-08-07 at 10 25 20

fernandomg:
@dennis00010011b that was quick! thanks for the catch, will review it

fernandomg:
Ok, so after reviewing the different scenarios that will arise with this PR and have a chat with @vbaranov. It was decided to close this PR and create a new solution for #1067, which is https://github.com/poanetwork/token-wizard/issues/1067#issuecomment-411397561

PR #1107, opened 2018-08-23
(Fix) check address of Proxy smart-contract for validity

### LINKED ISSUE #1042: (Fix) check address of Proxy smart-contract for validity
**Problem**: we don't check the validity of input in case of proxy in querystring (`?addr=` param)

**Solution**: 
Add such check in [getAddr](https://github.com/poanetwork/token-wizard/blob/2.0/src/utils/utils.js#L21) function:
1. as a first step we should check, that this is a valid address with `web3.utils.isAddress`
2. as a bonus we could check, that this is a Proxy-like contract: it has not empty property `app_exec_id `
3. as a triple bonus we could check, that `exec-id` of this Proxy smart-contract exists in Auth-os registry: we could request `getAdmin` method of `...idx` smart-contract (`MintedCappedIdx` or `DutchIdx`) with gotten exec-id from previous step.

mariano-aguero:
Closes #1042


mariano-aguero:
@dennis00010011b 

> Crash after invalidCrowdsaleProxyAlert

Done

Issue #1038, opened 2018-07-13
(Feature) add ABI-encoded parameters of Proxy smart contract deployment to Publish page


vbaranov:
child of #1035 

**Solution**: ABI-encoded parameters of Proxy smart contract deployment could be added below of the source-code of Proxy smart-contract (#1037)

PR #1115, opened 2018-08-28
(Feature) add source code of Proxy smart-contract to Publish page

### LINKED ISSUE #1037: (Feature) add source code of Proxy smart-contract to Publish page
child of #1035 

**Solution**: The source code of Proxy smart-contract could be added in the textarea below all other properties of crowdsale, as it is done in `master` branch of Token Wizard (live link: https://wizard.poa.network/)

fernandomg:
Closes #1037 

I've just added sources for `.abi`, `.bin` and `.src`.

### DIFF
--- a/src/components/stepFour/index.js
+++ b/src/components/stepFour/index.js
@@ -547,7 +547,6 @@ export class stepFour extends React.Component {
             {tiersSetupBlock}
             {this.renderContractSource('src')}
             {this.renderContractSource('abi')}
-            {this.renderContractSource('bin')}
           </div>
         </div>
         <div className="button-container">
--- a/src/components/stepFour/index.js
+++ b/src/components/stepFour/index.js
@@ -17,7 +17,8 @@ import {
   TOAST,
   CROWDSALE_STRATEGIES_DISPLAYNAMES,
   TEXT_FIELDS,
-  PUBLISH_DESCRIPTION
+  PUBLISH_DESCRIPTION,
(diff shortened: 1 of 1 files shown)


dennis00010011b:
@fernandomg 
Publish page: section `CROWDSALE PROXY CONTRACT CREATION CODE` contains **contract byte code**  (it matches ./public/contracts/MintedCappedProxy.bin in case of Minted crowdsale)

Should contains **contract creation code = contract byte code + constructor parameters**

fernandomg:
@dennis00010011b Mh... It was intended to be the 'Contract Creation Code' for the proxy contract. And so far it looks that it's not necessary on the TW.

Also to find the 'Contract Creation Code' for an already deployed contract, I have to find the original `tx` when the Proxy contract was deployed, and gather the information from there (@dennis00010011b, @vbaranov please correct me if I'm wrong).

So I just leave only the `.abi` and `.src` content in the Publish page.

vbaranov:
@fernandomg we need to give source code and ABI encoded parameters to a user. By having these set of parameters (with contract name and compiler version) a user will be able to verify Proxy smart-contract and will see ABI of it after verification. Addition of ABI encoded parameters is requested in #1038. Addition of source code is implemented here. Display of ABI is redundant. Let's remove it from Publish page.

fernandomg:
@vbaranov thanks for the clarification, change done.

Issue #1084, opened 2018-08-14
Apply new design to Landing Page


pablofullana:
Child of #1082

Issue #1083, opened 2018-08-14
Add assets to support new design


pablofullana:
Child of #1082

gabitoesmiapodo:
This could be tested, but it should be merged into some temporary branch, like "2.0-styles" where we continually merge the latest 2.0 updates, I think.

I don't think that extracting and adding all the new assets is the best thing to do at this step, so I'm doing that screen by screen, component by component, unless somebody needs some asset to do their job.

How to test:

- There should be some new folders and files, preparing the terrain for the new styles, check that those exist:

token-wizard/src/assets/stylesheets/v2.0 contains some new folders, files, and explanations.

token-wizard/src/assets/images/v2.0 has some folders and images for the new styles

This structure will allow us to have the old and new styles in parallel, minimizing the risk of losing something currently working. Once the work is finished, we'll delete the old files and that's it.

- It's possible to switch styles using: 

http://dappurl/?uiversion=2 (v2, currently ugly and broken)
http://dappurl/?uiversion=1 (old working version)
http://dappurl/  a mix of versions 1 and to, so as not to compensate whatever styles are not done yet.

This should work.

When we finish we will also delete the old styles.

gabitoesmiapodo:
PR at https://github.com/poanetwork/token-wizard/pull/1119

PR #1119, opened 2018-08-30
Feature/#1083 2.0 restyling

gabitoesmiapodo:
Pull request to temporary branch _origin/integration/#1082-new-design_ where we can continually merge the latest "new styling" working branches (#1083, #1084, etc.), and the 2.0 PRs, so as to keep everything as updated as possible.

As for the #1083 tasks: I don't think that extracting and adding all the new assets is the best thing to do at this step, so I will be doing that screen by screen / component by component, unless somebody needs some asset to do their job.

**How to test:**

1- There should be some new folders and files, preparing the terrain for the new styles, check that those exist:

_token-wizard/src/assets/stylesheets/v2.0_ contains some new folders, files, and explanations.
_token-wizard/src/assets/images/v2.0_ has some folders and images for the new styles

This structure will allow us to have the old and new styles in parallel, minimizing the risk of losing something currently working. Once the work is finished, we'll delete the old files and that's it.

2- It's possible to switch styles using:

http://dappurl/?uiversion=2 (v2, currently ugly and broken)
http://dappurl/?uiversion=1 (v1 working version)
http://dappurl/ a mix of versions 1 and 2, to compensate for whatever styles are not done yet.

This should work.

When we finish we will also delete the old styles.


vbaranov:
@dennis00010011b could we in parallel with restyling update, also updating e2e tests to support new interface?

dennis00010011b:
@vbaranov 
e2e test for this feature in branch `2.0-newDesign`:  https://github.com/poanetwork/token-wizard-test-automation/tree/2.0-newDesign

PR #1133, opened 2018-09-10
(Fix) Tests for Home Component

### LINKED ISSUE #1129: (Test) Home Component test case for 'stepOne' navigation
**Problem:** When testing `Home` component (see PR #1122), the method `getCrowdsaleAssets` fails with the following error.

```
(node:28260) UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'URL' of null
    at new XMLHttpRequest (/token-wizard/node_modules/jsdom/lib/jsdom/living/xmlhttprequest.js:116:39)
    at /token-wizard/src/utils/utils.js:81:5054
    at new Promise (<anonymous>)
    at fetchFile (/token-wizard/src/utils/utils.js:81:4900)
    at setFlatFileContentToState (/token-wizard/src/utils/utils.js:81:4778)
    at _callee2$ (/token-wizard/src/stores/utils.js:60:129)
    at tryCatch (/token-wizard/node_modules/regenerator-runtime/runtime.js:65:40)
```

**Solution:** Review and refactor to prevent the `Unhandled PromiseRejectionWarning` message.

**Note:** There's a commented out test case that should pass if this issue is fixed (`should navigate to StepOne`)

fernandomg:
Closes #1129

### DIFF
--- a/src/stores/utils.js
+++ b/src/stores/utils.js
@@ -56,7 +56,7 @@ async function getCrowdsaleAsset(contractName, stateProp, networkID) {
       ? setFlatFileContentToState(`./contracts/${stateProp}.sol`)
       : Promise.resolve()
   const whenBin =
-    stateProp === 'MintedCappedProxy' || stateProp === 'DutchProxy' || stateProp === 'ProxiesRegistry'
+    stateProp === 'MintedCappedProxy' || stateProp === 'DutchProxy'
       ? setFlatFileContentToState(`./contracts/${stateProp}.bin`)
       : Promise.resolve()
   let abi
@@ -2541,7 +2541,10 @@ async function getCrowdsaleAsset(contractName, stateProp, networkID) {
       abi = []
       break
   }
-  const addr = contractName ? JSON.parse(process.env[`${REACT_PREFIX}${contractName}_ADDRESS`] || {})[networkID] : null
+
+  const addr = contractName
+    ? JSON.parse(process.env[`${REACT_PREFIX}${contractName}_ADDRESS`] || '{}')[networkID]
+    : null
 
   logger.log(abi, addr)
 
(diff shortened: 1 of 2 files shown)


src/stores/utils.js
```diff
@@ -56,7 +56,7 @@ async function getCrowdsaleAsset(contractName, stateProp, networkID) {
       ? setFlatFileContentToState(`./contracts/${stateProp}.sol`)
       : Promise.resolve()
   const whenBin =
-    stateProp === 'MintedCappedProxy' || stateProp === 'DutchProxy' || stateProp === 'ProxiesRegistry'
+    stateProp === 'MintedCappedProxy' || stateProp === 'DutchProxy'
```
vbaranov: @fernandomg could you remind, why do we need to remove `ProxiesRegistry` from here?

fernandomg: If you serve the app running `npm run start` it will try to find the `ProxiesRegistry` file and will fail, but instead of throwing a `404` error, it returns the `index.html` page content... and it's stored in the `CrowdsaleStore` as the `ProxiesRegistry`'s `bin`.

image

While if you build the app (`npm run build`) and serve it from within the `/build` directory with `npx http-server` you'll be able to get what I assume we will get from netlify server (a `404` error). And the page won't continue.

image

---
As the app worked fine with the wrong `bin` for `ProxiesRegistry` I assumed that it's not required, besides we're not storing it in the `/contracts` directory.

If it will be required in the future, we can add it by that time. Plus we can properly handle the `404` and inform the user about the issue with a modal/toaster. I can create an issue for the last if you think it's necessary.

vbaranov: @fernandomg you're right, we do not need to check `ProxiesRegistry` here. Thank you for the clarification. And yes, your proposal to properly handle `404` does make sense to me.

PR #1197, opened 2018-10-23
(Fix) Incorrect build sequence

vbaranov:
The sequence of subtasks of `build` script is incorrect:
it copies everything to `build` folder and only then generates metadata for contracts. The reverse sequence should be. This PR fixes the issue.


mariano-aguero:
@vbaranov 

I found this 
screenshot_20181023_101328

State off build folder
screenshot_20181023_104340


Fixed with the last commit
screenshot_20181023_104252

State of build folder
screenshot_20181023_104849

The rest look good to me,

Regards

PR #1204, opened 2018-10-29
(Feature) Apply new design to step 4: "Publish"

### LINKED ISSUE #1088: Apply new design to Step 4 - Publish
Child of #1082

gabitoesmiapodo:
Closes #1088 

**Description:**

Here's the implementation of Step 4's styles.

There are some changes to the original designs (fields' order, added TxStatus elements) in the hopes of making everything clearer and easier to the end user.

**Screenshots:**

screen shot 2018-10-29 at 15 18 37
screen shot 2018-10-29 at 16 16 56
screen shot 2018-10-29 at 15 25 41
screen shot 2018-10-29 at 16 26 16
screen shot 2018-10-29 at 15 25 51
screen shot 2018-10-29 at 16 26 27


dennis00010011b:
@gabitoesmiapodo 
Multitiers crowdsale: only Tier1 column is displayed in modal. 
screen shot 2018-10-30 at 10 11 24 am

dennis00010011b:
@gabitoesmiapodo 
Extra title, shouldn't be here
screen shot 2018-10-30 at 10 41 15 am

dennis00010011b:
@gabitoesmiapodo 
Typo in Step3 
screen shot 2018-10-30 at 11 00 14 am

gabitoesmiapodo:
@vbaranov

1- column with tx status is not fixed. It is moving depending on the length of the current status message. Could we fix it? -> The "Tx Name" column's width is now fixed, so no columns should move now. I don't know if there's a better fix for this, other than changing how this information is shown.

2- There is no response on copy action. Could we add a tooltip? -> Yeah, now a message is shown when the user clicks on this.

3- Text in one line -> Sure. I think there was a mistake and some text got repeated.

4- Text in one line for wallet addres -> Sure, but for "Whitelist with cap" only, because there's a column for "Burn excess" next to in "Dutch Auction".

@dennis00010011b

- Multitiers crowdsale: only Tier1 column is displayed in modal. -> I think only first tier and tiers with a whitelist are displayed (maybe someone can confirm this?)

dennis00010011b:
@gabitoesmiapodo 
1. Yes, you are right
> * Multitiers crowdsale: only Tier1 column is displayed in modal. -> I think only first tier and tiers with a whitelist are displayed (maybe someone can confirm this?)


2.There is no modal in Step 4 after this commit https://github.com/poanetwork/token-wizard/pull/1204/commits/6ba57375cfbbe8f1c3958981d9fb445b102f06c3

screen shot 2018-10-31 at 7 11 21 pm

gabitoesmiapodo:
@dennis00010011b 

`1- Extra title, shouldn't be here`

Removed.

`2- Typo in Step3`

Fixed.

```
3- Dutch: tier 's info is placed in chaotic order. Would be better:

Start time      ---- End Time
Global min cap  ---- max cap
Min rate        ---- max rate
Enable whitelisting
```

Fixed.

`4- There is no modal in Step 4 after this commit 6ba5737`

Yeah, now it's back.

dennis00010011b:
@gabitoesmiapodo 
LGTM, except two things:
1. I still cannot compile build on MacOS unless to rename folder  from `src/components/crowdsale` to src/components/Crowdsale`
2. Should be mentioned  about all tiers in modal? Example, crowdsale has 3 tiers: 1&3 without whitelist, 2nd whiteist, as a result the Tier3 isn't displayed. Would be better to group  no-whitelisted tiers together?

screen shot 2018-11-01 at 3 49 21 pm

gabitoesmiapodo:
@dennis00010011b 

> I still cannot compile build on MacOS unless to rename folder from src/components/crowdsale to src/components/Crowdsale

I used another way to change the files / folders' names on MacOS and have them recognized by Git, please have a look now.

screen_shot_2018-11-02_at_08_35_35

> Should be mentioned about all tiers in modal? Example, crowdsale has 3 tiers: 1&3 without whitelist, 2nd whiteist, as a result the Tier3 isn't displayed. Would be better to group no-whitelisted tiers together?

We should ask @vbaranov or maybe @fernandomg  about this. I'm not particularly fond about the way this information is displayed now, for several reasons: It's not clear that only tiers with a whitelist are included, tiers' names aren't displayed, I can't cancel or leave this section if something goes wrong (or if I just want to), and generally speaking It's not always clear what's going on (to me at least).

Maybe we should just add a short text clarifying that only tiers' with a whitelist are included here? That would be a small improvement, but I'm not sure if that's really the problem.

I created an issue with a proposal of what I think would be a better way to display all this information to the user: https://github.com/poanetwork/token-wizard/issues/1208

fernandomg:
@gabitoesmiapodo I just added a fix to eliminate the duplicated in-progress button

image

The rest looks fantastic.

